PR #2, opened 2020-06-24
docs: fix pin link description of measurement availability on an edge

jdbruijn:
During the recent import I noted this difference in the docs and what actually happens.


everhardt:
Great. If we'd ever require it, we could decide to add optional parameters `asOf: Date` and `until: Date` to be able to create pinLinks in the past or future.

PR #6, opened 2020-11-09
Initialize anomaly detection branch

z841860053:
Add new routes and models for anomaly detection

### DIFF
--- a/src/models/measurement.ts
+++ b/src/models/measurement.ts
@@ -37,6 +37,10 @@ const schema = (apiVersion: number): Joi.ObjectSchema => {
       .description('The measured value is significand * 10 ^ orderOfMagnitude. It has as many significant figures as the significand has (except when the significand is 0, then the number of significant figures is not defined)'),
     performance: Joi.number().integer().default(0)
       .description('0: within thresholds, 1: outside serious thresholds but inside critical thresholds, 2: outside critical thresholds'),
+    anomalousDetector: Joi.number().integer().valid(0, 1, null).default(null)
+      .description('Anomaly detector classification, 0: not an anomaly, 1: is an anomaly, null: unknown'),
+    anomalousUser: Joi.number().integer().valid(0, 1, null).default(null)
+      .description('User classification, 0: not an anomaly, 1: is an anomaly, null: unknown'),
   });
 };
 
@@ -49,6 +53,8 @@ interface Measurement {
   orderOfMagnitude: number;
   significand: number;
   performance: number;
+  anomalousDetector?: 0 | 1 | null;
+  anomalousUser ?: 0 | 1 | null;
 }
 
 interface MeasurementV1 {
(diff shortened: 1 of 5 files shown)


src/models/measurement.ts
```diff
@@ -37,6 +37,10 @@ const schema = (apiVersion: number): Joi.ObjectSchema => {
       .description('The measured value is significand * 10 ^ orderOfMagnitude. It has as many significant figures as the significand has (except when the significand is 0, then the number of significant figures is not defined)'),
     performance: Joi.number().integer().default(0)
       .description('0: within thresholds, 1: outside serious thresholds but inside critical thresholds, 2: outside critical thresholds'),
+    anomalousDetector: Joi.number().integer().valid(0, 1, null).default(null)
+      .description('Anomaly detector classification, 0: not an anomaly, 1: is an anomaly, null: unknown'),
+    anomalousUser: Joi.number().integer().valid(0, 1, null).default(null)
+      .description('User classification, 0: not an anomaly, 1: is an anomaly, null: unknown'),
```
everhardt, line 43: Sorry Jichen, you couldn't have known, but in Joi we actually do use boolean. So it becomes `Joi.boolean().allow(null).default(null)`.

src/models/measurement.ts
```diff
@@ -49,6 +53,8 @@ interface Measurement {
   orderOfMagnitude: number;
   significand: number;
   performance: number;
+  anomalousDetector?: 0 | 1 | null;
+  anomalousUser ?: 0 | 1 | null;
```
everhardt, line 57: These keys are non-optional, as they are either provided by the backend or by the default in Joi. So the `?` should go in both lines.

src/routes/measurement/setAnomalousUser.ts
```diff
@@ -0,0 +1,35 @@
...
+
+const controllerGeneratorOptions: ControllerGeneratorOptions = {
+  method: 'post',
+  path: '/:hashId/anomalousUser',
+  params: Joi.object().keys({
+    hashId: Joi.string().required().example('qoa978'),
+  }).required(),
+  body: Joi.object().keys({
+    anomalousUser: Joi.number().integer().valid(0, 1, null),
+    clientHashId: Joi.string().required().example('qoa978'),
```
everhardt, line 24: See discussion in platform repo about `clientHashId`

everhardt (changes requested):

src/routes/measurement/set-anomalous-user.ts
```diff
@@ -0,0 +1,33 @@
+import Joi from 'joi';
+import { ControllerGeneratorOptions } from '../../comms/controller';
+
+interface Request {
+  params: {
+    hashId: string;
+  };
+  body: {
+    anomalousUser: boolean | null;
```
everhardt, line 9: this null should probably go now?

everhardt (changes requested):

PR #9, opened 2021-01-07
added defaultOrderOfMagnitude field to the Quantity model and api

yarsky-tgz:


src/routes/quantity/add.ts
```diff
@@ -18,6 +19,10 @@ const controllerGeneratorOptions: ControllerGeneratorOptions = {
   body: Joi.object().keys({
     name: Joi.string().required().example('Temperature'),
     unit: Joi.string().required().example('K').description('Will be displayed with an SI-prefix (eg. k or M) if relevant'),
+    defaultOrderOfMagnitude: Joi.number().integer().min(-128).max(127)
+      .required()
```
everhardt, line 23: We have to assume there will be clients on the current API for quite some time:
* days with our own platform-client UI, if they do not refresh the page and ignore the popup
* weeks or months for the current machine-to-machine implementations of our API

Therefore, I think we need to change `.required()` to `.default(0)`. The TypeScript interface can remain as it is, as (for the server) that describes the state of the incoming body after it has been validated by Joi.

everhardt (changes requested):

everhardt (approved):

PR #10, opened 2021-01-10
feat: default thresholds for quantities

yarsky-tgz:
added default thresholds into quantity interface and route schemas for both environments

### DIFF
--- a/src/routes/quantity/update.ts
+++ b/src/routes/quantity/update.ts
@@ -10,10 +10,10 @@ interface Request {
     name?: string;
     unit?: string;
     defaultOrderOfMagnitude?: number;
-    defaultCriticallyLowThreshold: SiNumber | null;
-    defaultLowThreshold: SiNumber | null;
-    defaultHighThreshold: SiNumber | null;
-    defaultCriticallyHighThreshold: SiNumber | null;
+    defaultCriticallyLowThreshold?: SiNumber | null;
+    defaultLowThreshold?: SiNumber | null;
+    defaultHighThreshold?: SiNumber | null;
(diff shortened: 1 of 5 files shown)


src/routes/quantity/update.ts
```diff
@@ -9,6 +10,10 @@ interface Request {
     name?: string;
     unit?: string;
     defaultOrderOfMagnitude?: number;
+    defaultCriticallyLowThreshold: SiNumber | null;
+    defaultLowThreshold: SiNumber | null;
+    defaultHighThreshold: SiNumber | null;
+    defaultCriticallyHighThreshold: SiNumber | null;
```
everhardt, line 16: Shouldn't these be optional?

src/routes/report-type/update.ts
```diff
@@ -4,11 +4,16 @@ import {
...
+import { schema as siNumberSchema, SiNumber } from '../../models/si-number';
 
 type RequestQuantity = {
   name: string;
   unit: string;
   defaultOrderOfMagnitude: number;
+  defaultCriticallyLowThreshold: SiNumber | null;
+  defaultLowThreshold: SiNumber | null;
+  defaultHighThreshold: SiNumber | null;
+  defaultCriticallyHighThreshold: SiNumber | null;
```
everhardt, line 16: Shouldn't these be optional?

everhardt (changes requested):
Almost approved it without any comment, but found one in the end ;). Thanks Yaroslav, like the progress you make!

everhardt (approved):

PR #11, opened 2021-01-11
feat: default thresholds for quantities

yarsky-tgz:
removed optional modifier from default thresholds


everhardt (changes requested):
Good one! The Joi schema should also be changed I think, as it now does not have a `.required()` or `.default(..)`. I think it should have a `.default(null)`, to not introduce a breaking change for existing API consumers. While you're at it: I think `defaultOrderOfMagnitude` could use a default as well..

everhardt (approved):

PR #14, opened 2021-02-01
Feat platform 563 create export request by measurement filter hashid

yarsky-tgz:
* added name, description fields to MeasurementFilterContent interface, added MeasurementFilterContentByHashId interface to settings.add-export route
* reusing contentSchemaAlternatives

### DIFF
--- a/src/routes/settings/add-export.ts
+++ b/src/routes/settings/add-export.ts
@@ -1,6 +1,6 @@
 import Joi from 'joi';
 import { ControllerGeneratorOptions } from '../../comms/controller';
-import { AllContent, contentSchema } from '../../models/export-request';
+import { AllContent, contentSchemaAlternatives } from '../../models/export-request';
 
 interface MeasurementFilterContentByHashId {
   type: 'measurementFilter';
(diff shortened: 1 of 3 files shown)


src/routes/settings/add-export.ts
```diff
@@ -18,7 +37,13 @@ const controllerGeneratorOptions: ControllerGeneratorOptions = {
   method: 'post',
   path: '/export',
   body: Joi.object().keys({
-    content: contentSchema.required(),
+    content: Joi.alternatives().try(
+      Joi.object().keys({
+        type: Joi.string().required().valid('measurementFilter').example('measurementFilter'),
+        measurementFilterHashId: Joi.string().required(),
+      }).required(),
+      ...contentSchemaAlternatives,
```
everhardt, line 45: This introduces a breaking change, as the second element of `contentSchemaAlternatives` has a new required prop: `name`. I like name to be required in the response of settings/find-export-request.ts, so breaking the link here and just repeating the existing schema in this file would be acceptable to me. It's deprecated anyway, so also easier to clean up afterwards.

everhardt (changes requested):

src/routes/settings/add-export.ts
everhardt: My comment might have been multi-interpretable, but I actually meant it the other way around:
* the response of routes/settings/find-export-request.ts should link to models/export-request (models are used almost exclusively in responses, almost never in requests)
* the request of routes/settings/add-export.ts should not link to models/export-request but have its own definition the of  `body.content` property (without the name property in there for a measurementFilter-type content, as that would be breaking).

yarsky-tgz: Yep, misunderstanding happened here, sorry. My thoughts gone completely wrong way, reverted that changes, done as you asking and in place were you've asked this time :)

everhardt (approved):
Merci!

PR #16, opened 2021-02-10
feat: increase quantities count

yarsky-tgz:
increased allowed count of quantityHashIds at related schemas


everhardt (approved):

everhardt:
I've just released version 16.12.2

PR #18, opened 2021-02-18
feat: added `charts` fields into DeviceType routes and model

yarsky-tgz:


src/models/device-type.ts
```diff
@@ -13,7 +13,22 @@ const baseSchema = Joi.object().keys({
     pinFieldConfigurations: fieldConfigurationsFromServerSchema.required()
       .description('Defines deviceFields on the pin the channel is connected to. Can be used in report type functions. See the chapter on open fields on how to use this'),
     defaultPinName: Joi.string().example('Anode').description('If undefined, the channel cannot be linked to a pin'),
+    charts: Joi.array().items(Joi.object().keys({
```
everhardt, line 16: As we'll only implement this client side after the server implements this change: shall we make both `models.deviceType.charts` and `models.deviceType.channels[].charts` required and non-null? Saves the front-end the hassle of checking that every time

src/routes/device-type/add.ts
```diff
@@ -12,7 +12,22 @@ interface Request {
...
+      }[];
     }[];
+    charts: {
+      title: string | null;
+      series: {
+        channelIndex: number;
+        quantityHashId: string;
+        color: string;
+      }[];
+    }[] | null;
```
everhardt, line 30: I don't think the `null` is needed: client can just set `[]`. I would like to make this object optional though, for backwards compatability.

src/routes/device-type/update.ts
```diff
@@ -18,7 +18,22 @@ interface Request {
...
+      }[];
     }[];
+    charts?: {
+      title: string | null;
+      series: {
+        channelIndex: number;
+        quantityHashId: string;
+        color: string;
+      }[];
+    }[] | null;
```
everhardt, line 36: null is not needed I think (see above)

everhardt (changes requested):

src/routes/device-type/add.ts
everhardt: Both charts objects in deviceType.add should be optional for backwards compatibility.

src/routes/device-type/update.ts
everhardt: Both charts objects in deviceType.add should be optional for backwards compatibility (the one on root-level already is optional, so that's good!) .

yarsky-tgz:
@everhardt updated, please check

src/routes/device-type/add.ts
```diff
@@ -37,7 +52,22 @@ const controllerGeneratorOptions: ControllerGeneratorOptions = {
       pinFieldConfigurations: fieldConfigurationsToServerSchema.required()
         .description('Defines deviceFields on the pin the channel is connected to. Can be used in report type functions. See the chapter on open fields on how to use this'),
       defaultPinName: Joi.string().example('Anode').description('If undefined, the channel cannot be linked to a pin'),
+      charts: Joi.array().items(Joi.object().keys({
+        title: Joi.string().allow(null).example('Red wire charts').required(),
+        series: Joi.array().items(Joi.object().keys({
+          quantityHashId: Joi.string().example('x18a92').required(),
+          color: Joi.string().example('#ff00ff').required(),
+        })),
```
everhardt, line 60: Line 60 should get a `.required()` I think, to be in line with the typescript Interface (which is correct).

everhardt (changes requested):
Sorry Yaroslav, found one small error I think ;).

src/routes/device-type/add.ts
yarsky-tgz: oh, fixed )

everhardt (approved):
Merci!

PR #19, opened 2021-02-25
feat(graph.findPinQuantities):

yarsky-tgz:
added `graph.findPinQuantities` route

### DIFF
--- a/src/routes/graph/find-pin-quantities.ts
+++ b/src/routes/graph/find-pin-quantities.ts
@@ -1,0 +1,43 @@
+import Joi from 'joi';
+import { ControllerGeneratorOptions } from '../../comms/controller';
+
+import { schema as quantitySchema, Quantity } from '../../models/quantity';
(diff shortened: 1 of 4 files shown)


everhardt (changes requested):
Looks good Yaroslav!

* Can you rename the `hashId` parameter to `pinHashId`? It's not immediately clear to me what kind of hashId it is as we mention both pin and quantity in the route. Also, could you update the example with the hashId of the pin model, `e13d57`?
* This is totally my bad (I asked for `findPinQuantities`), but I think we should rename the route to `getPinQuantities` as all `find..` routes implement the `TableQuery` interface and use `tableQuerySchemaGenerator`. Basically, this route does not do paging, so `get` is better than `find`. The `rows: ResponseRow` structure is then not required anymore (you can rename it if you like something else than `rows` better, although `rows` is still fine by me).
* Could you also update `src/routes/graph/routes.ts` please?

everhardt:
Great, thanks!

everhardt:
Just released this in v16.16.0

PR #20, opened 2021-02-28
feat(graph.setThreshold):

yarsky-tgz:
request body `thresholds` property can be null


everhardt (changes requested):
Can you make response.hashId optional? With a description `'If thresholds is set to null, this key will not be present'`

everhardt (approved):

everhardt:
Great!

everhardt:
Released in v16.17.0

PR #21, opened 2021-03-02
feat(device-type,quantity,chart): quantity color

yarsky-tgz:
* added `color` to the `Quantity` model and routes
 * made `series[].color` nullabe everywhere at charts

### DIFF
--- a/src/routes/quantity/add.ts
+++ b/src/routes/quantity/add.ts
@@ -5,6 +5,7 @@ import { schema as siNumberSchema, SiNumber } from '../../models/si-number';
 interface Request {
   body: {
     name: string;
+    color?: string;
     unit: string;
     defaultOrderOfMagnitude: number;
     defaultCriticallyLowThreshold: SiNumber | null;
@@ -24,6 +25,7 @@ const controllerGeneratorOptions: ControllerGeneratorOptions = {
   path: '/',
   body: Joi.object().keys({
     name: Joi.string().required().example('Temperature'),
+    color: Joi.string().default('#ff00ff').example('#ff00ff'),
     unit: Joi.string().required().example('K').description('Will be displayed with an SI-prefix (eg. k or M) if relevant'),
     defaultOrderOfMagnitude: Joi.number().integer().min(-128).max(127)
       .default(0)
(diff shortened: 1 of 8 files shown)


src/routes/quantity/add.ts
```diff
@@ -5,6 +5,7 @@ import { schema as siNumberSchema, SiNumber } from '../../models/si-number';
 interface Request {
   body: {
     name: string;
+    color?: string;
```
everhardt, line 8: It's a bit ambiguous for who we provide the interface: if it's for the backend, color is not optional as the default is used. If it's for the front-end, it should indeed be optional. I propose to aim it at the backend: that uses it right now and it only requires typescript front-ends using this version or newer to explicitly specify a color.

src/routes/quantity/update.ts
```diff
@@ -8,6 +8,7 @@ interface Request {
   };
   body: {
     name?: string;
+    color?: string;
```
everhardt, line 11: Same 'o

src/routes/quantity/add.ts
everhardt: Using the same rationale you might argue that a `.default(..)` in any model (in src/models, which are only used in responses from the server) will never be used, as the Interface provides it correctly as non-optional.

everhardt (changes requested):

src/routes/quantity/add.ts
yarsky-tgz: I was thinking about back compatibility here, but have been mistaken with layer a bit and almost added redundant check to controller :)

default value already handles old clients.

so I've removed `?`

src/routes/quantity/update.ts
yarsky-tgz: but this is `update`, each field is optional here, or may be I do not understand you completely ?

In `add` I've removed `?`

everhardt: Sorry, you are absolutely right!

everhardt (approved):

everhardt:
Released in 16.18.0

PR #23, opened 2021-03-05
fix(graph/set-threshold): right types for Response.hashId

yarsky-tgz:
forgotten to change response


everhardt (approved):

everhardt:
Released in 16.20.0

PR #28, opened 2021-03-25
feat: added device quantities to quantity.add, quantity.find, quantity.update

yarsky-tgz:


everhardt:
Nice! I think you missed one:

> add deviceQuantities: Quantity[] to the Response (and Joi schema of course) of quantity.get. When called by a supplier environment, this key is always []. In a monitoring environment, it is a list of mapped supplier quantities.

(for full disclosure: I just renamed it to `deviceQuantities`, it used to be called `linkedSupplierQuantities`).

everhardt:
Sorry, never mind that: it was already in an earlier commit, so let's leave it as it is!

PR #29, opened 2021-03-29
feat: removed setReportingFrequency supplier activity and all measurementCycles mentions, except get-pin-group < v3

yarsky-tgz:


jdbruijn:
The `expectNextReportBefore` is now implemented everywhere, so conflicts here can be resolved and this can be merged as far as I'm concerned.

everhardt:
Missed that it contains breaking changes, fixed that separately in https://github.com/withthegrid/platform-sdk/commit/06155c268d52ac2863456972c085d1a78bf889ce

PR #32, opened 2021-04-28
Feat 421 audit log generic

KZavorotna:
Related PRs:
https://github.com/withthegrid/platform-client/pull/494
https://github.com/withthegrid/platform/pull/669

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,10 +16,10 @@
   "scripts": {
     "start": "npm run watch-ts",
     "build": "npm run build:esm && npm run build:cjs",
-    "build:cjs": "rm -rf cjs/ && tsc -p tsconfig.cjs.json",
-    "build:esm": "rm -rf esm/ && tsc",
+    "build:cjs": "rimraf -rf cjs/ && tsc -p tsconfig.cjs.json",
+    "build:esm": "rimraf -rf esm/ && tsc",
     "prepare": "npm run build",
-    "watch-ts": "rm -rf esm/ && tsc -w",
+    "watch-ts": "rimraf -rf esm/ && tsc -w",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -29,7 +29,8 @@
     "@types/qs": "^6.9.6",
     "axios": "^0.21.1",
     "joi": "^17.4.0",
-    "qs": "^6.10.1"
+    "qs": "^6.10.1",
+    "rimraf": "^3.0.2"
```
everhardt, line 33: Can you put rimraf in the devDependencies? No need to have it in the implementing repo's (platform-client for example) I think?

src/models/log.ts
```diff
@@ -3,8 +3,8 @@ import Joi from 'joi';
 const schema = Joi.object().keys({
   hashId: Joi.string().required().example('op09a'),
   userHashId: Joi.string().allow(null).required().example('b45zo0'),
-  objectType: Joi.string().allow(null).required().example('command'),
-  objectHashId: Joi.string().allow(null).required().example('ga9741s'),
+  objectType: Joi.string().allow(null).example('command'),
+  objectHashId: Joi.string().allow(null).example('ga9741s'),
```
everhardt, line 7: This model is only used for responses from the server. I think those will always contain these two properties, so they will still have to be required

everhardt (changes requested):

package.json
KZavorotna: Updated

src/models/log.ts
KZavorotna: Updated

KZavorotna:
PR is updated

PR #34, opened 2021-04-30
feat: extended BaseFieldConfiguration interface and schema

yarsky-tgz:


src/models/fields/base-field-configuration.ts
```diff
@@ -1,36 +1,96 @@
...
 }
 
 interface BaseFieldConfiguration {
   key: string;
-  name: string;
-  inputType?: 'text' | 'textarea' | 'select' | 'radio' | 'switch' | 'checkbox' | 'file' | 'files';
+  type: 'string' | 'boolean' | 'number' | 'integer';
+  name: string | Translations;
+  inputType: 'text' | 'textarea' | 'select' | 'radio' | 'switch' | 'checkbox' | 'file' | 'files';
+  defaultValue?: string | boolean | number;
```
everhardt, line 85: The problem with default values, is that for the client, the keys are optional, while for the backend they are always present. I propose to use the TS Interface for the backend, so with `defaultValue` and `valueOptions` being required. A client that is using TS will then not have the benefit of default values (ie. he has to supply values himself), but I think that outweighs the extra checks the backend has to do.

src/models/fields/base-field-configuration.ts
```diff
@@ -1,36 +1,96 @@
...
+    .when(Joi.ref('upperbound'), {
+      is: Joi.number().required(),
+      then: Joi.string().valid('number').default('number'),
+    }),
+  name: stringOrTranslationsSchema.required(),
+  inputType: Joi.string()
+    .when(Joi.ref('valueOptions'), {
+      is: Joi.array().required(),
+      then: Joi.string()
+        .valid('select')
```
everhardt, line 26: Shouldn't this be `.valid('select', 'radio')`?

src/models/fields/base-field-configuration.ts
```diff
@@ -1,36 +1,96 @@
...
+        .default('select'),
+      otherwise: Joi.string()
+        .valid('text', 'textarea', 'select', 'radio', 'switch', 'checkbox', 'file', 'files')
```
everhardt, line 29: I think `select` and `radio` should not be in this list?

src/models/fields/base-field-configuration.ts
```diff
@@ -1,36 +1,96 @@
...
-  })).allow(null)
-    .description('If null, inputType should not be select or radio. If not null, input type should be select or radio'),
-  prefix: Joi.string().description('Not available for inputTypes \'radio\', \'switch\', \'checkbox\', \'file\' and \'files\''),
-  suffix: Joi.string().description('Not available for inputTypes \'radio\', \'switch\', \'checkbox\', \'file\' and \'files\''),
+  })).allow(null).default(null)
+    .description('If null, inputType should not be select or radio. If not null, input type should be select or radio. Default: null'),
+  regex: Joi.string().description('If provided, type should be string and the provided value should adhere to this regex'),
+  lowerbound: Joi.number().description('If provided, type should be number or integer (and lowerbound should be integer as well in case type is integer) and provided value should not be lower than this value'),
+  upperbound: Joi.number().description('If provided, type should be number or integer (and lowerbound should be integer as well in case type is integer) and provided value should not be higher than this value'),
```
everhardt, line 58: Could you drop the `(and lowerbound should be integer as well in case type is integer)` part in the description? I know I wrote it myself, but I think it's clearer without.

everhardt (changes requested):

everhardt:
Thank you!

PR #35, opened 2021-05-03
feat: analytics route

everhardt:
part of withthegrid/platform#572

### DIFF
--- a/src/models/analytics-query.ts
+++ b/src/models/analytics-query.ts
@@ -1,0 +1,85 @@
+import Joi from 'joi';
+
+const constraintSchema = Joi.object().keys({
+  field: Joi.alternatives().try(Joi.string(), Joi.number()).required(),
+  comparison: Joi.string().required(),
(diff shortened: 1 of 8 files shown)


src/models/analytics-query.ts
```diff
@@ -0,0 +1,85 @@
...
+
+type TimeGranularity =
+  'day'
+  | 'week'
+  | 'month'
+  | 'year'
+  | 'dayOfTheWeek'
+  | 'monthOfTheYear';
+
+type Comparison = '=' | '>=' | '<=' | '>' | '<' | 'like' | '<>' | 'isNull' | 'isNotNull';
```
yarsky-tgz, line 44: i think should be also added `notLike` comparison

yarsky-tgz (commented):
.

src/models/analytics-query.ts
everhardt: Yeah, good point. I won't add it now (as the SDK has already progressed a few versions further, for which the platform has not yet caught up), but will put it on my todo!

PR #36, opened 2021-05-06
feat: put back pin links (temporary)

yarsky-tgz:
I've removed pin-links models to fast, we need it yet for frontend, to switch without problems

### DIFF
--- a/src/models/index.ts
+++ b/src/models/index.ts
@@ -35,6 +35,7 @@ import * as fieldsToServerFull from './fields/fields-to-server-full';
 import * as fieldsToServerUpdate from './fields/fields-to-server-update';
 import * as updatableFieldConfigurations from './fields/updatable-field-configurations';
 import * as pin from './pin';
+import * as pinLink from './pin-link';
 import * as quantity from './quantity';
 import * as webRequest from './web-request';
 import * as siNumber from './si-number';
@@ -100,6 +101,7 @@ export {
   fieldsToServerUpdate,
   updatableFieldConfigurations,
   pin,
+  pinLink,
   quantity,
   siNumber,
   supplier,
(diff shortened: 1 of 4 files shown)


everhardt (approved):

everhardt:
I see you just added "WIP", so I'll hold off merging it.

yarsky-tgz:
> I see you just added "WIP", so I'll hold off merging it.

one problem here: 16.32.0 is released. 

Can you please release these change as 16.31.1, without changes from 16.32.0?

then may be merging even not needed

everhardt:
I'm afraid the only way is up, both because of CI and NPM/ GitHub registries.

yarsky-tgz:
If talking about npm registries - i was doing such trick sometimes:

https://stackoverflow.com/questions/24691314/npm-publish-patch-for-earlier-major-version

Can you do `npm publish` manually?

everhardt:
I'd rather do two extra commits and keep everything in line instead of creating an exception now with unknown consequences. What will be the problem if I just merge this pull on top of main?

yarsky-tgz:
From my previous experience (I was patching some package many times, both 1.x and 2.x versions) only problem is that you need to add some 'fake' tag (not to overwrite `latest`), but seems it can be removed easily, by command `npm dist-tag rm <pkg_name> <tag>`.

The only risk of this operation is to overwrite the `latest` tag, but it's hard to do, because npm allows that only with `--force` flag.

Many packages offers patches for previous major versions, so it seems to be usual and safe action. 

The reason I want to do so: I just do not want to switch from unfinished task to another and loose concentration.

yarsky-tgz:
another option is to add into this PR revert of 16.32.0

everhardt:
Alright. We can't do it on the main branch, as that will rewrite history and fuck up the CI, but I can create a new branch off 16.31.0 (say "16.61.x") and direct this PR there. Then I can manually publish that version to the GH registry (which is what our own repo's are using)

PR #37, opened 2021-05-06
feat: added edge link fields to pin Joi schema

yarsky-tgz:


everhardt:
Great! Merging it, as it is backwards compatible:
* [x] A client that is not updated can communicate with a server that has been updated
* [x] A client that is updated can communicate with a server that not yet been updated

PR #38, opened 2021-05-06
feat: remove pin link model

yarsky-tgz:
This reverts commit de5a6c19abaf00c157fd4efcc7437b62717ec35c.

part of withthegrid/platform#504


everhardt (approved):
I missed this last time I reviewed it, but this is a breaking change: it will break clients using older sdk versions. 

AFAIK there are no clients in production other than platform-client that use these two endpoints, so fine to merge this PR as long as the client is updated before we update the server. I would prefer to therefore hold off the merge until the client PR is as good as ready, do you agree?

yarsky-tgz:
Yes, completely :)

yarsky-tgz:
Removed old methods for pin links editing also

everhardt (approved):
Looks good! It has "WIP", can I already merge it?

yarsky-tgz:
> Looks good! It has "WIP", can I already merge it?

Yes, removed WIP

PR #39, opened 2021-05-06
feat: gridName in graph.addPinGroup and graph.updatePinGroup

everhardt:
part of withthegrid/platform#683

### DIFF
--- a/src/routes/graph/add-pin-group.ts
+++ b/src/routes/graph/add-pin-group.ts
@@ -14,6 +14,7 @@ interface Request {
     fields: FieldsToServerFull;
     mapLayer?: string;
     gridHashIds?: string[];
+    gridName?: string;
     photo?: string;
   };
 }
@@ -65,6 +66,7 @@ const controllerGeneratorOptions: ControllerGeneratorOptions = {
     }
     return baseBody.keys({
       gridHashIds: Joi.array().items(Joi.string()).description('PinGroups will be added at the end of the list in a grid'),
+      gridName: Joi.string().description('Ensures the pinGroup is part of a grid with the provided name. If multiple grids exist with the same name, one is chosen at random'),
     }).required();
   },
   right: { environment: 'STATIC' },
--- a/src/routes/graph/update-pin-group.ts
+++ b/src/routes/graph/update-pin-group.ts
@@ -17,6 +17,7 @@ interface Request {
     } | null;
     fields?: FieldsToServerUpdate;
     gridHashIds?: string[];
+    gridName?: string;
     mapLayer?: string;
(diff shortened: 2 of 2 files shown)


everhardt:
We could! It's not a problem in practice right now, so it wouldn't be a high prio thing to address. And to make it guaranteed would require locking the table (as a unique index on it isn't possible given that we also store deleted grids).

yarsky-tgz:
> We could! It's not a problem in practice right now, so it wouldn't be a high prio thing to address. And to make it guaranteed would require locking the table (as a unique index on it isn't possible given that we also store deleted grids).

Regarding unique index:  we can to add into index deleted_at field itself then. 

What do you think?

everhardt:
That's not possible I think, see https://dev.mysql.com/doc/refman/5.7/en/create-index.html: 

> A UNIQUE index permits multiple NULL values for columns that can contain NULL.

yarsky-tgz:
Oh, you're right! But I've googled and found some interesting recipe, how to make it possible just with specific table structure:

```sql
CREATE TABLE `things` (
  `id` int(11) unsigned NOT NULL AUTO_INCREMENT,
  `name` varchar(255) NOT NULL,
  `deleted_at` datetime DEFAULT NULL,
  `is_not_deleted` int(1) GENERATED ALWAYS AS (if(isnull(`deleted_at`),1,NULL)) VIRTUAL,
  PRIMARY KEY (`id`),
  UNIQUE KEY `name` (`name`,`is_not_deleted`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
```

and it works:

image

yarsky-tgz:
And more information about how indexing of such GENERATED ... VIRTUAL  columns works:

https://dev.mysql.com/doc/refman/5.7/en/create-table-secondary-indexes.html

And that means that JSON can be indexed then too! 
Second part of article mentioned above is telling in details about that.

yarsky-tgz:
Such index helper fields can be hidden then by using VIEW. So we will permit accidentally re-generation of each value on SELECT (not good for any query), what means it will be calculated only on writes and stored to index only. 

If such fields will be really needed to access from query I prefer then even to have them STORED.

everhardt:
Haha, we used to have such a construction for device identifiers! Worked great to avoid duplicates, but caused some headaches when importing/ exporting data and with MySQL clients that couldn't deal with it (eg. https://github.com/sequelpro/sequelpro/issues/2200). 

If the problem of non-unique grids would have been bigger/ more important I would have agreed to the solution you propose, as it is technically sound. I think though that the issue is too small to warrant spending much more time on, so would propose to keep it as it is.

yarsky-tgz (approved):

PR #42, opened 2021-05-11
feat: add edge hash id to pin add update routes

yarsky-tgz:


src/routes/graph/update-pin.ts
```diff
@@ -24,6 +25,7 @@ const controllerGeneratorOptions: ControllerGeneratorOptionsWithClient = {
   }).required(),
   body: Joi.object().keys({
     fields: fieldsToServerUpdateSchema.example({ id: 'My port' }),
+    edgeHashId: Joi.string().allow(null).default(null).example('ka08d'),
```
everhardt, line 28: This should not have a default I think?

yarsky-tgz: So in code It's typed `string | null`. Can be changed to `string | undefined` if you want to drop `default` from Joi.

Should I make such change for `edgeHashId` ?

everhardt (approved):

PR #43, opened 2021-05-11
#454 - Added notifications level for group

KZavorotna:
Related PRs:
https://github.com/withthegrid/platform/pull/687
https://github.com/withthegrid/platform-client/pull/505


everhardt:
Made two changes:
* `level` -> `notificationLevel` as per https://github.com/withthegrid/platform-client/issues/454#issuecomment-815865116. It makes it in line with getGrid. I know setNotifications has `level`, but there you can derive that it is about notifications from the name of the route
* made the joi schema required (as the TS interface is so as well)

PR #45, opened 2021-05-16
feat: extended base field configuration

yarsky-tgz:
Part of withthegrid/platform#664


yarsky-tgz:
It has some additions to be compared with previous PR, which has been reverted: https://github.com/withthegrid/platform-sdk/commit/d60b4b2210a37827bbe37eb25c15db0c3938662b

yarsky-tgz:
That was long journey :)

From beginning I was looking for solution with live schema based validation on frontend, and during investigation our typescript editor found, that it supports not only json sources, but has reach support of JSON Schema with all modern drafts for validation, providing autocomplete, document and many others. Abilities of it are comparable with Joi, and here we have example of it.

And I've run into new quest: how to provide that schema.

First decision was obvious: compile it from Joi, we have everything there. Even descriptions. 

But after looking at all solutions, I was scared about hackish way of their working, and way, how they supports conditions. 

So I've decided just to write that json by my own. During writing it I've found, why all that `joi-to-json-schema` and `joi-to-swagger`, `joi-to-json` cannot compile conditions: logic of their scope and behaviour is completely different.

Also I've found, that JSON Schema does not support inheritance, and here is official explanation why:

image

In short: if you try to do inheritance, you will have for example schema which requires and rejects the same field at the same time, so it cannot be valid in both cases: then field exists and then not.

But actually you can do whatever you want, while you are just copy and pasting things.
So my json schema started growing. First almost-full functional version was 500 lines long.
And it was awful for edit, view and work with it, and while trying another one hack to make it shorter I suddenly found [typescript-json-schema](https://github.com/YousefED/typescript-json-schema)

And it does all I want, but [`BaseFieldConfiguration`](https://github.com/withthegrid/platform-sdk/pull/45/files#diff-76185ae702a966ee07eb5716f0ac9655d6e462fdc3a29ed8988740155daab4feR128) type grown much.

BTW. I've tried to apply same logic as in TS through `Joi.alternatives()` and it just do not work. I've tried everything. So I've left our `when` based solution, and added some extensions to it, to be closer to TS.

yarsky-tgz:
And some preview, how it looks on frontend:

image

yarsky-tgz:
sizes of json schema, unminified json and gzipped:

image

a lot of copy & paste there, so it archives very well

src/models/fields/base-field-configuration.ts
```diff
@@ -1,38 +1,241 @@
...
-  valueOptions?: ValueOption[] | null;
-  prefix?: string;
-  suffix?: string;
-  hint?: string;
+  /**
+   * name of the field to be shown in UI
+   */
+  name: StringOrTranslations;
+  /**
+   * show this field in UI only if value of another form field, identified by provided `key`, is
```
everhardt, line 110: > of another form field

I propose to make that 'of an earlier form field', to rule out circular references. I know it's also possible to do more advanced analysis and allow later referencing form fields, but I also think it does not make sense in the UI that an earlier component is shown when you change something further down.

src/models/fields/base-field-configuration.ts
```diff
@@ -1,38 +1,241 @@
...
+import {
+  stringOrTranslationsSchema,
+  StringOrTranslations,
+} from '../../translations';
 
 const schema = Joi.object().keys({
   key: Joi.string().pattern(/^[a-z][a-zA-Z\d]*$/).required().example('id'),
-  name: Joi.string().required().example('ID as used in our geographic information system'),
-  inputType: Joi.string().valid('text', 'textarea', 'select', 'radio', 'switch', 'checkbox', 'file', 'files')
+  type: Joi.string().valid('string', 'boolean', 'number', 'integer')
```
everhardt, line 10: These `.when` constructs are a bit hard to validate, as only the first one that matches is evaluated. Why did you opt for this instead of for the `Joi.alternatives()` mechanism that matches the TS interface?

everhardt (commented):
Looks good! Mostly have clarifying questions.

src/translations.ts
```diff
@@ -0,0 +1,18 @@
...
+    nl: Joi.string().required(),
+  }),
+);
+
+export interface Translations {
+  en: string,
+  nl: string,
+}
+
+export type Locale = keyof Translations;
```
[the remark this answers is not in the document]
yarsky-tgz: Yeah, makes sense!

src/translations.ts
```diff
@@ -0,0 +1,18 @@
+import Joi from 'joi';
```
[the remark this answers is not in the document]
yarsky-tgz: Sure!

src/models/fields/field-configuration-from-udf.ts
```diff
@@ -10,7 +10,7 @@ const schema = baseFieldConfigurationSchema.keys({
   .tag('fieldConfigurationFromUdf')
   .description('Defines which data can be stored in form fields and how that configuration should be sent to the server.');
 
-interface FieldConfigurationFromUdf extends BaseFieldConfiguration {
+type FieldConfigurationFromUdf = BaseFieldConfiguration & {
```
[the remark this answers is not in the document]
yarsky-tgz: That's not benefit :) While we need `BaseFieldConfiguration` to be `Union` it cannot be and declared as `interface` but only as `type`. 

So we already cannot `extend` the type, but we can get same result type, while declaring `FieldConfigurationFromUdf` as `type` and `&` the additional fields to our `BaseFieldConfiguration` there. 

So, I was forced by TS to do above changes.

src/models/fields/base-field-configuration.ts
yarsky-tgz: I've tried! It is not working :( 

It throws error that it cannot find any alternative for such object (!). And shows that schema with applied defaults, taken from one of alternative schema. I've spent I think hours just on trying to make it work, rewriting that alternative schemas  again and again with same result.

yarsky-tgz: From terminal tab history

```
Body: "environment.fieldConfigurations.pinGroups[0]" does not match any of the allowed types (HTTP 500)
```
......
```
      "pinGroups": [
        {
          "key": "assetId",
          "hint": "",
          "name": "Meetpaalnummer",
          "valueOptions": null,
          "type": "string",
          "defaultValue": "",
          "inputType": "text"
        },
```

in DB it has only `key`, `name` and `hint` ! All these defaults taken from **alternative** schema, expected one to be applied! 

But - you see result on top of this comment.

yarsky-tgz: https://github.com/withthegrid/platform-sdk/blob/28554e66f460662e94a9c00fa94e412375988d11/src/models/fields/base-field-configuration.ts#L27

here is that Joi alternatives schema.

everhardt: If you print the validation error (using JSON.stringify), you'll get an extensive explanation about why it did not match. The relevant part of it:
```
{
  "message": "\"defaultValue\" is not allowed to be empty",
  "path": [
    "defaultValue"
  ],
  "type": "string.empty",
  "context": {
    "label": "defaultValue",
    "value": "",
    "key": "defaultValue"
  }
},
```

The issue is a very bad legacy decision by Joi to not allow an empty string by default. You have to add `.allow('')` after the `defaultValue` schema and then it validates like a charm :).

Note to self, it took me a while to get it running, but this is the solution: `ts-node --project ./tsconfig.cjs.json ./src/test`

yarsky-tgz: > you'll get an extensive explanation about why it did not match

Oh, thought that I have already everything possible at console :(

Great then! Will revert back to alternatives now!

yarsky-tgz:
Can be reviewed again & removed WIP :)

yarsky-tgz:
but let's wait with merging, I think

src/routes/environment/update.ts
```diff
@@ -18,7 +19,7 @@ interface Request {
       nodes: UpdatableFieldConfigurations;
       pins: UpdatableFieldConfigurations;
     };
-    locale?: 'nl' | 'en';
+    locale?: Locale;
```
everhardt, line 22: Shouldn't the Joi-schema also be replaced (with `localeSchema`)?

src/models/fields/base-field-configuration.ts
```diff
@@ -1,39 +1,358 @@
...
+   */
+  defaultValue: StringOrTranslations;
+  /**
+   * @default null
+   */
+  valueOptions: null | [];
+  /**
+   * @default "text"
+   */
+  inputType: 'text' | 'textarea' | 'translations';
```
everhardt, line 191: What's the  `'translations'` type?

src/models/fields/base-field-configuration.ts
```diff
@@ -1,39 +1,358 @@
...
+    inputType: Joi.string().valid('text', 'textarea').default('text'),
+    regex: Joi.string(),
+    ...prefixesMixin,
+  }).required(),
+  commonSchema.keys({
+    type: Joi.string().valid('string').default('string'),
+    defaultValue: stringOrTranslationsSchema.allow('').default(''),
+    valueOptions: Joi.array().items(Joi.object().keys({
+      text: stringOrTranslationsSchema.required(),
+      value: Joi.string().required().description('Will be passed through parser'),
```
everhardt, line 43: The `Will be passed through parser` comment has become irrelevant I think?

everhardt (changes requested):
Looks good Yaroslav, only have very minor remarks

src/models/fields/base-field-configuration.ts
yarsky-tgz: I've added such input type to display `translation-text-field` in `editable-field `, and thought it would be nice to add it here as well.

everhardt: I don't think that works: a filled out translation-text-field would produce data that does not conform to `BaseField` but is an object or array itself. 

(it's not in the Joi schema by the way).

src/routes/environment/update.ts
yarsky-tgz: Sure, found few more places to apply it

src/models/fields/base-field-configuration.ts
yarsky-tgz: Fixed schema, added `Translations` to `BaseField` (including Joi schema)

src/models/fields/base-field-configuration.ts
yarsky-tgz: Removed

src/models/fields/base-field-configuration.ts
everhardt: Yeah, but that means that we'll be storing complexer objects in the `fields` columns in the database, while right now they are always objects that are one level deep. I'm only for increasing complexity if the benefit is clear. I don't really see a good use case for this, what do you have in mind?

yarsky-tgz: I was thinking so: string value can be constant (identifier, serial number, hash, etc) or it can be translateable (name, description). 

Now I've reviewed few production environments and see that all actual fields string values do not need to be translated.

So I agree that users do not really need it and will remove it now from `BaseFieldConfiguration` and remove `Translations` from `BaseField` union.

everhardt (approved):
Shall I merge it already or hold off a bit?

yarsky-tgz:
> Shall I merge it already or hold off a bit?

Now it can be :)

PR #48, opened 2021-06-03
feat: strict show if in field configuration

yarsky-tgz:
Part of withthegrid/platform#664


everhardt:
What issue does it solve? Looks fine by the way

yarsky-tgz:
> What issue does it solve? Looks fine by the way

If `value` is a string, but completely numeric, it's converted to Number:

for example:

```
    first: Joi.alternatives().try(
      Joi.number(),
      Joi.string(),
    ),
    second: Joi.alternatives().try(
      Joi.number(),
      Joi.string(),
    ),
  });

  console.log(schema.validate({ first: '1984', second: '1984aa' }));
```

produces output:

```
{ value: { first: 1984, second: '1984aa' } }
```

And with `strict()` schema is denied to convert type.

PR #58, opened 2021-07-24
feat: support for expressions in analytics

yarsky-tgz:
Part of withthegrid/platform#678


src/models/analytics-query.ts
```diff
@@ -59,15 +67,15 @@ interface Condition {
...
-  | { type: 'mean'; condition?: Condition; field: string } // SUM(CASE WHEN condition THEN field ELSE 0 END) / SUM(CASE WHEN condition/1 THEN 1 ELSE 0 END)
-  | { type: 'min'; condition?: Condition; field: string } // MIN(CASE WHEN condition THEN field ELSE NULL END)
-  | { type: 'max'; condition?: Condition; field: string } // MAX(CASE WHEN condition THEN field ELSE NULL END)
+  | { type: 'sum'; condition?: Condition; field: Field } // SUM(CASE WHEN condition THEN field ELSE 0 END)
+  | { type: 'mean'; condition?: Condition; field: Field } // SUM(CASE WHEN condition THEN field ELSE 0 END) / SUM(CASE WHEN condition/1 THEN 1 ELSE 0 END)
+  | { type: 'min'; condition?: Condition; field: Field } // MIN(CASE WHEN condition THEN field ELSE NULL END)
+  | { type: 'max'; condition?: Condition; field: Field } // MAX(CASE WHEN condition THEN field ELSE NULL END)
   ;
 
 type TimeGroupColumn = { type: 'timeGroup', field: string; granularity: TimeGranularity };
```
everhardt, line 76: Did you intentionally not change `field: string` to `field: Field` in `TimeGroupColumn`? If so, I think I agree, as we cannot easily track server side whether that expression can actually be time grouped.

yarsky-tgz: Sorry for late response. 

As I see currently we just cannot make time group by expression, while date needed for time grouping and we expect for number from calculation. 

Also I've limited field-picker for expressions on frontend to provide number-only fields, because arithmetic operations can be applied only to numbers in MySQL.

PS. From other side we can convert dates to timestamps for calculations and convert expression result back to date for grouping by time, but not sure it will be transparent and actually needed by users.

everhardt: Alright, we're in line. Let's leave it as it is, we can always expand the functionality to include dates.

PR #60, opened 2021-07-28
feat: added supplierOnly option to the base field configuration

yarsky-tgz:
Part of withthegrid/platform#517


everhardt:
See https://github.com/withthegrid/platform/issues/517#issuecomment-892928036

PR #62, opened 2021-08-05
#453 subscribe to all issues

KZavorotna:
Related to: 
https://github.com/withthegrid/platform/pull/743
https://github.com/withthegrid/platform-sdk-private/pull/9
https://github.com/withthegrid/platform-client/pull/593


src/routes/environment/find.ts
```diff
@@ -34,6 +35,9 @@ const controllerGeneratorOptions: ControllerGeneratorOptionsWithoutClientOrSuppl
       environment: environmentSchema.required(),
       environmentRights: Joi.array().items(Joi.string()).required().example(['STATIC', 'USERS'])
         .description('See the getting started section about rights'),
+      notificationLevel: Joi.number().allow(null).required()
```
everhardt, line 38: This should be an exception to my remark that "everywhere where `environmentRights` are supplied in a response, also supply `notificationLevel: 0 | 1 | 2 | null`". The `environmentRights` here are not about the user performing the request, but the user of the row and I don't think it makes sense that the requesting user knows about that. So can you drop `notificationLevel` again here?

src/routes/authentication/machine-login.ts
```diff
@@ -15,6 +15,7 @@ interface Response {
   user: User;
   environment?: Environment;
   environmentRights?: string[];
+  notificationLevel: 0 | 1 | 2 | null;
```
everhardt, line 18: Can it be made optional without any dire consequences, and the Joi schema as well?

src/routes/settings/get.ts
```diff
@@ -11,6 +11,7 @@ interface Response {
   environment?: Environment;
   environmentRights?: string[];
   user: User;
+  notificationLevel: 0 | 1 | 2 | null;
```
everhardt, line 14: Can it be made optional without any dire consequences, and the Joi schema as well?

KZavorotna:
Ready for review

everhardt:
Thanks! Merged this into https://github.com/withthegrid/platform-sdk/commit/d4d65f2c61e70f415b45309cdf7a71513d47b03e, so will close this PR

PR #63, opened 2021-08-05
feat: added allowNull property to BaseFieldConfiguration

yarsky-tgz:
Part of withthegrid/platform-client#562

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6706,11 +6706,6 @@
         "safer-buffer": "^2.0.2",
         "tweetnacl": "~0.14.0"
       },
-      "bin": {
-        "sshpk-conv": "bin/sshpk-conv",
-        "sshpk-sign": "bin/sshpk-sign",
-        "sshpk-verify": "bin/sshpk-verify"
-      },
       "engines": {
         "node": ">=0.10.0"
       }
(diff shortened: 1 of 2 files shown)


yarsky-tgz:
It was interesting challenge to make properties mutually exclusive.

Joi has `.xor()` method on objects, which allows to do same thing, that I've got in TS with `RequireOnlyOne` type, but `.xor()` disallows (indeed) usage of `.default()` on affected properties. We need to set `.default()` for `defaultValue` on old records from database, that's why I've used alternatives and field configurations megaschema grown again a bit. :)

yarsky-tgz:
Oh, after touching UI I've got that's wrong solution. Challenge continues. :)

jdbruijn:
FWIW the mutually exclusive part isn't a must have from my side, I was just thinking it might be easier code-wise and for users creating and writing form-fields. If it saves a lot of trouble code-wise and it makes it easier to follow, I'm also with having both the `allowNull` and `default` options. I didn't have a strong case for either one or the other.

everhardt:
> Challenge continues. :)

If the challenge still continues, I'd argue for "Let's keep it simple" and let the Joi/ TS interface be more lenient in return for some simpler code. If the challenge is done and it works (so I should review), then let's leave it as it is!

yarsky-tgz:
Challenge is done, can be reviewed!

> I'd argue for "Let's keep it simple" 

It just became bigger, cannot say it started to be more complicated. Just more alternatives where needed. 
Schema and interface are bit huge already, but precise.

PR #64, opened 2021-09-01
feat: removed settings.request-export-download and download.get route…

yarsky-tgz:
…s, added downloadUrl field to ExportRequest model

part of withthegrid/platform#356


everhardt (approved):
Great, thanks!

PR #66, opened 2021-09-21
feat: add next page offset

yarsky-tgz:
part of withthegrid/platform#768


everhardt (approved):
Looks good to me! I'll hold of merging until https://github.com/withthegrid/platform/pull/792#pullrequestreview-759612785 is also approved.

PR #68, opened 2021-10-05
feat: left constraint can be expression

yarsky-tgz:
part of withthegrid/platform#678

### DIFF
--- a/src/models/index.ts
+++ b/src/models/index.ts
@@ -24,6 +24,7 @@ import * as issue from './issue';
 import * as issueComment from './issue-comment';
 import * as label from './label';
 import * as log from './log';
+import * as locale from './locale';
 import * as mapLayer from './map-layer';
 import * as measurement from './measurement';
 import * as measurementFilter from './measurement-filter';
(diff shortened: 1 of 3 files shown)


src/models/index.ts
```diff
@@ -24,6 +24,7 @@ import * as issue from './issue';
 import * as issueComment from './issue-comment';
 import * as label from './label';
 import * as log from './log';
+import * as locale from './locale';
```
everhardt, line 27: I dropped locale on purpose, as this index is used for generating the documentation and `locale` is not an object which the doc generator requires for each model. Could you please revert that? Or are you missing the reference to it anywhere?

yarsky-tgz: I am still having 230 lint errors on `platform` with fresh `main`, because of missing `models.locale` declarations, so I've put it back app to launch locally. 

Is part of this update for `platform` released already?

everhardt: Just discussed through chat: we'll accept this change and fix the documentation repo instead.

PR #73, opened 2021-11-07
fix: allowNull is not required

yarsky-tgz:
if defaultValue is set.

Fixes withthegrid/platform-client#724


everhardt (changes requested):
shouldn't it be `allowNull?: false` in these cases? Otherwise stored values that have `allowNull: false` become invalid

yarsky-tgz:
True! Fixed.

PR #75, opened 2021-11-25
#664 - Added default graph range for environment

KZavorotna:
Is required for https://github.com/withthegrid/platform-client/pull/746

Is related to https://github.com/withthegrid/platform-client/pull/746

### DIFF
--- a/src/models/environment.ts
+++ b/src/models/environment.ts
@@ -48,7 +48,7 @@ interface Environment {
   locale: Locale;
   expiresAt: Date | null;
   createdAt: Date;
-  defaultGraphRange: string;
+  defaultGraphRange: string | null;
 }
 
 export { schema, Environment };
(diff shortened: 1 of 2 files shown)


src/models/environment.ts
```diff
@@ -25,6 +25,7 @@ const schema = Joi.object().keys({
   locale: localeSchema.required().example('en'),
   expiresAt: Joi.date().allow(null).required().example(null),
   createdAt: Joi.date().required().example('2019-12-31T15:23Z'),
+  defaultGraphRange: Joi.string().allow(null).required().example('30d'),
```
yarsky-tgz, line 28: is `allow(null)` needed? It's not reflected on type.

yarsky-tgz (changes requested):
One small comment

src/models/environment.ts
KZavorotna: Updated

yarsky-tgz (approved):

everhardt (approved):
Looks good! Will hold off merging until platform and platform-client implementations are ready

PR #77, opened 2021-12-03
#712 - Added hd button for graphs

KZavorotna:
Is required for https://github.com/withthegrid/platform-client/pull/762


src/routes/chart/get.ts
```diff
@@ -33,6 +34,7 @@ const controllerGeneratorOptions: ControllerGeneratorOptionsWithClient = {
       pinHashId: Joi.string().required().example('e13d57'),
       quantityHashId: Joi.string().required().example('sajia1'),
     })).required(),
+    highResolution: Joi.boolean().required(),
```
everhardt, line 37: Can you replace `.required()` with `.default(false)`? That we we ensure backwards compatibility (unrefreshed browsers and people using our sdk directly will not send highResolution).

yarsky-tgz (approved):

src/routes/chart/get.ts
KZavorotna: Updated

everhardt (approved):

PR #80, opened 2021-12-30
feat: add subtitle to analytics widget

parkerjohn:
Is required for [#804](https://github.com/withthegrid/platform-client/pull/804)

Is required for [#887](https://github.com/withthegrid/platform/pull/887)

### DIFF
--- a/src/models/analytics-panel.ts
+++ b/src/models/analytics-panel.ts
@@ -59,7 +59,7 @@ const schema = Joi.object().keys({
   layout: layoutSchema.default([]),
   cards: Joi.array().items(Joi.object().keys({
     title: Joi.string().required().example('My widget'),
-    subTitle: Joi.string().allow('').required().example('Widget subtitle'),
+    subTitle: Joi.string().example('Widget subtitle'),
     query: analyticsQuerySchema.required(),
     visualisation: analyticsVisualisationSchema.required(),
   })).required(),
--- a/src/models/analytics-panel.ts
+++ b/src/models/analytics-panel.ts
@@ -59,7 +59,7 @@ const schema = Joi.object().keys({
   layout: layoutSchema.default([]),
   cards: Joi.array().items(Joi.object().keys({
     title: Joi.string().required().example('My widget'),
-    subTitle: Joi.string().allow('').optional(),
+    subtitle: Joi.string().allow('').example('My widget description'),
     query: analyticsQuerySchema.required(),
     visualisation: analyticsVisualisationSchema.required(),
   })).required(),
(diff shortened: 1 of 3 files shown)


src/models/analytics-panel.ts
```diff
@@ -59,6 +59,7 @@ const schema = Joi.object().keys({
   layout: layoutSchema.default([]),
   cards: Joi.array().items(Joi.object().keys({
     title: Joi.string().required().example('My widget'),
+    subTitle: Joi.string().allow('').optional(),
```
everhardt, line 62: * Shall we write it as `subtitle`? Wikipedia writes it as one word: https://en.wikipedia.org/wiki/Subtitles
* `.optional()` is not needed: if you do not specify `.required()` it is optional by default

src/models/analytics-panel.ts
```diff
@@ -81,6 +82,7 @@ interface AnalyticsPanel {
   layout: AnalyticsPanelLayout;
   cards: {
     title: string;
+    subTitle: string;
```
everhardt, line 85: As it's optional, it should be `subtitle?: string`

src/routes/analytics/add-panel.ts
```diff
@@ -32,6 +32,7 @@ const controllerGeneratorOptions: ControllerGeneratorOptionsWithClient = {
     layout: layoutSchema.default([]),
     cards: Joi.array().items(Joi.object().keys({
       title: Joi.string().required().example('My widget'),
+      subTitle: Joi.string().allow('').optional(),
```
everhardt, line 35: It's missing from the TS interface above.

everhardt (changes requested):

src/models/analytics-panel.ts
yarsky-tgz, line 62: shouldn't be better to provide also any example?

yarsky-tgz, line 62: actually `.allow('')` is redundant, while field is optional

yarsky-tgz (changes requested):
Having some comments

src/models/analytics-panel.ts
everhardt: I don't think that's true:
image

yarsky-tgz: Oh. My bad, closing, sorry for disturbing.

parkerjohn: No idea what I was thinking there with the camelcase :)
Removed optional() everywhere

src/models/analytics-panel.ts
parkerjohn: Fixed

src/models/analytics-panel.ts
parkerjohn: Done

src/routes/analytics/add-panel.ts
parkerjohn: Added

everhardt (approved):
Ok! Will hold off merging until the platform-client PR is ready

yarsky-tgz (approved):

PR #82, opened 2022-01-04
feat(routes): suggestions.find

yarsky-tgz:
Part of withthegrid/platform-client#770

### DIFF
--- a/src/routes/suggestions/find.ts
+++ b/src/routes/suggestions/find.ts
@@ -26,9 +26,7 @@ interface Request {
   };
 }
 
-type EffectiveRequest = {
-  query: Required<Request['query']>;
-}
+type EffectiveRequest = Request;
 
 interface Response {
   results: string[];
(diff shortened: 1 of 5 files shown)


src/routes/suggestions/find.ts
```diff
@@ -0,0 +1,70 @@
...
+    search: string;
+    type: SuggestionType;
+    rowsPerPage: number;
+    offset?: string;
+  };
+}
+
+type EffectiveRequest = {
+  query: Required<Request['query']>;
+}
```
everhardt, line 31: I think offset is still optional in the EffectiveRequest, isn't it?

everhardt (changes requested):
One minor thing..

src/routes/suggestions/find.ts
yarsky-tgz: updated

everhardt (approved):

PR #83, opened 2022-01-17
feat: drop defaults for radio, switch and checkbox field configs and drop allowNull

everhardt:
* dropped `allowNull` from the TS Interface and only use it for backwards compatibility in the Joi schema. New logic (with the same result): 
  * no value (`undefined`) for a key is always allowed
  * if a default value is available it is _displayed_ instead (but not stored at the server)
* simplified base-field-configuration a bit, which also helps typescript in making sense of the object
* dropped pins from the  device.link route (which is a breaking change, but not one that affects external parties)

part of withthegrid/platform#912


yarsky-tgz (changes requested):
All comments that you've removed are not just comments, but instructions to JSON Schema, which is generated from type.

I think `minItems` at least should be left and defaults as well, may be.

And by the way, have you checked how updated JSON schema works at JSON editor in app?

everhardt:
Good point. Fix unfortunately is quite heavy, as it means introducing a new api version. I've also added the `minItems` back as that is very useful. The added value of `default` is quite limited, the monaco editor itself is already pretty smart in autocompleting.

yarsky-tgz (approved):
Great! Fields configuration types much more readable now :)

PR #91, opened 2022-01-21
feat: only show x last measurements per port per quantity

srieding:
Part of withthegrid/platform#863

I think I've seen that backwards compatible changes in this repository do not update the version?

### DIFF
--- a/src/models/analytics-query.ts
+++ b/src/models/analytics-query.ts
@@ -14,7 +14,7 @@ const constraintSchema = Joi.object().keys({
   }),
 });
 
-const limiterSchema = Joi.object().keys({
+const limitSchema = Joi.object().keys({
   limitedTo: Joi.number().required(),
   per: Joi.array().items(Joi.object().keys({
     columnIndex: Joi.number().required(),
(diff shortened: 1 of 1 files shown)


yarsky-tgz (approved):

everhardt (approved):
Looks good! Will merge when the platform PR is ready to merge.

src/models/analytics-query.ts
```diff
@@ -41,6 +52,7 @@ const schema = Joi.object().keys({
     }).required(),
   )).required(),
   filter: conditionSchema,
+  limiter: limiterSchema,
```
everhardt, line 55: While I went through https://github.com/withthegrid/platform/pull/936 I noticed that the wording is different from what we discussed earlier: `limiter` instead of `limitBy`. Why do you think the former is more fitting? It sounds more active than the other properties.

srieding: it is consistent with the filter which is not called filterBy

everhardt: Fair enough, but wouldn't it then be `limit`?

srieding: Is an option I considered. For me limiter would be an object that allows you to choose a limit and then limits for you. So to me it would be analogous to the machine that you use to configure the limit in your car. You can increase the limit set on the limiter, decrease the limit set on the limiter and the limiter will then make sure you will not drive faster than the limit configured. However if you strongly prefer limit or limitBy I'm happy to change it.

everhardt: For me limiter is the same kind of "machine" you describe. My only issue is that the other parameter names are not such machines.

everhardt (approved):

PR #108, opened 2022-02-14
fix 944 device table shows monitoring environments with attached devices

srieding:
is required by: https://github.com/withthegrid/platform/pull/1004
refs: withthegrid/platform#944

### DIFF
--- a/src/routes/suggestions/find.ts
+++ b/src/routes/suggestions/find.ts
@@ -1,8 +1,5 @@
 import Joi from 'joi';
-import {
-  ControllerGeneratorOptionsWithClient,
-  ControllerGeneratorOptionsWithoutClientOrSupplier, ControllerGeneratorOptionsWithSupplier,
-} from '../../comms/controller';
+import { ControllerGeneratorOptionsWithClientAndSupplier } from '../../comms/controller';
 
 const suggestionTypes = [
   'clientReportType.name',
@@ -32,16 +29,11 @@ interface Request {
 
 type EffectiveRequest = Request;
 
-interface ResponseRow {
-  hashId: string,
-    name: string,
-}
-
 interface Response {
-  results: ResponseRow[];
+  results: string[];
 }
 
-const controllerGeneratorOptions: ControllerGeneratorOptionsWithSupplier = {
+const controllerGeneratorOptions: ControllerGeneratorOptionsWithClientAndSupplier = {
   method: 'get',
   path: '/',
   query: Joi.object().keys({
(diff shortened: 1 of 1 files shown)


src/routes/suggestions/find.ts
```diff
@@ -28,11 +32,16 @@ interface Request {
...
 type EffectiveRequest = Request;
 
+interface ResponseRow {
+  hashId: string,
+    name: string,
+}
+
 interface Response {
-  results: string[];
+  results: ResponseRow[];
```
yarsky-tgz, line 41: `suggestions.find` route is done for working with `combobox` UI element. So It's expected always to return array of raw strings, because `combobox` is working so (without any identifiers and additional information regarding strings).

So I think this line should be not changed at all and vote `suggestion.find` to return only strings, as before.

And I think you need to find way to filter that values on server or use another route (`environment.find`), which already gives all model properties.

For more information regarding how `suggestions.find` were added you can look here: https://github.com/withthegrid/platform-client/issues/770

yarsky-tgz (changes requested):

src/routes/suggestions/find.ts
srieding: The choice to use the suggestion route was kind of made in the ticket before I started. But I do agree that making it work within suggestions is a bit clunky. @everhardt  where do you prefer it if you take this comment into account?

yarsky-tgz: > The choice to use the suggestion route was kind of made in the ticket before I started

And it sounds:

> do not filter on hashId, but on name and use the suggestions.find

so in any case no hashids were expected in response.

srieding: Ok, that's fair but environment name is not guaranteed to be unique.

everhardt: Sorry, I think I archived the email that informed me I was tagged here, without having given an answer. 

Right now, `device.find` can have a `search` parameter in the query string that can contain `environment: ${hashId}`. If we were to go for the solution I propose, this would change to `environment: ${name}`, which in the backend would be inserted in the query using ` LIKE ?` and `${name}%` or so.

Then, using suggestions.find could be used, as uniqueness is not relevant I think?

src/routes/suggestions/find.ts
```diff
@@ -1,5 +1,8 @@
 import Joi from 'joi';
-import { ControllerGeneratorOptionsWithClient } from '../../comms/controller';
+import {
+  ControllerGeneratorOptionsWithClient,
+  ControllerGeneratorOptionsWithoutClientOrSupplier, ControllerGeneratorOptionsWithSupplier,
```
everhardt, line 4: This now contains imports that are unused. Given the change we expect, I think `ControllerGeneratorOptionsWithClientAndSupplier` should be there :).

everhardt (changes requested):

everhardt (approved):
Looks good! Awaiting https://github.com/withthegrid/platform/pull/1004 before merging it.

PR #114, opened 2022-02-15
feat: initialValue for switch, radio and checkbox form field configs

everhardt:
part of https://github.com/withthegrid/platform/issues/1014


srieding (approved):
It is pragmatic. I hope we will not run into more problems in this direction making us regret that we don't save the initial value for other fields.

yarsky-tgz:
Interesting move in that pretty long default & optional values story :)

PR #126, opened 2022-03-04
feat(issue-trigger-rules): add model and controller

Arinono:
## Authors
@Arinono 

## Issues
withthegrid/platform#704

### Model

An `IssueTriggerRule` is composed of:
- `priorityLevel`: 1 or 2 right now
- `missedReports`: the number of missed reports to trigger the issue
- `offlineForSeconds`: the number of seconds since the device was last online to trigger the issue

On top of that, the actual schema also has two fields:
- `clientHashId`: to override issue trigger rules on an environment level
- `deviceTypeHashId`: (optional) to override issue trigger rules on a device type level

### Controllers

They just follow HTTP CRUD basics.
One thing I'm not sure about are the `getBy(Client|DeviceType)` endpoints. They are useful, and I don't think we need a `find` for this resource, so I went with this.
Feel free to give your opinion or suggestions about this.

### DeviceType level override

We will see in the platform that I made sure the DB and models/controllers are "ready" for `deviceType` level overriding. But it doesn't mean that the `client` and `monitor-connectivity` will implement this from the get go. But it felt easy enough to design the table and models with this in mind instead of iterating on this later.
So not YAGNI, but I really think `deviceType` level overriding will come handy.
All that said, right now, I haven't worked on the UI yet, so maybe I'll implement it at the end of the day.

### Rights

I went with what made sense from what I've seen around, but I don't know if I put the good rights on the endpoints.

### ClientHashId

If we look at the `add`, for example, the `body` doesn't asks for the `clientHashId` because when I did the implementation in the platform. I used the existing `req.userClient.object.clientId`. I don't know if it's a desired pattern or not.


src/routes/issue-trigger-rule/get.ts
```diff
@@ -0,0 +1,40 @@
...
+    hashId: Joi.string().required().example('xd2rd4'),
+  }),
+  right: { environment: 'READ' },
+  response: Joi.object().keys({
+    deviceTypeHashId: Joi.string().example('xd2rd4'),
+    priorityLevel: Joi.number().integer().valid(1, 2).example('1'),
+    missedReports: Joi.number().integer().required().example('12'),
+    offlineForSeconds: Joi.number().integer().required().example('1800'),
+  }),
+  description: 'Returns all issue trigger rules for an environment',
```
srieding, line 32: Looks like this should state: Get a issue trigger rule identified by its hashId

src/routes/issue-trigger-rule/getByClient.ts
```diff
@@ -0,0 +1,35 @@
+import Joi from 'joi';
```
srieding, line 1: I think naming this file find and then using: 
  method: 'get',
  path: '/',
  query: etc
would be more consistent with the rest of the project

src/routes/issue-trigger-rule/getByDeviceType.ts
```diff
@@ -0,0 +1,37 @@
+import Joi from 'joi';
```
srieding, line 1: see comment above. I think you could do without this file by making the previous one more generic

srieding (changes requested):

src/routes/issue-trigger-rule/get.ts
srieding: Do you dislike the get a specific xxxx by its hashId descriptions that we have on other get routes?

src/routes/issue-trigger-rule/get.ts
```diff
@@ -0,0 +1,39 @@
...
+  }),
+  right: { environment: 'READ' },
+  response: Joi.object().keys({
+    clientHashId: Joi.string().example('xd2rd4').required(),
+    deviceTypeHashId: Joi.string().example('xd2rd4'),
+    priorityLevel: Joi.number().integer().valid(1, 2).example('1'),
+    missedReports: Joi.number().integer().required().example('12'),
+    offlineForSeconds: Joi.number().integer().required().example('1800'),
+  }),
+  description: 'Gets all issue trigger rules for an environment',
```
srieding, line 31: I would still prefer the description that we use elsewhere: 'Get a specific xxxxx identified by its hashId'

srieding (approved):

src/routes/issue-trigger-rule/get.ts
Arinono: No, I forgot about this one, updated after your review

srieding: cool

src/routes/issue-trigger-rule/update.ts
```diff
@@ -0,0 +1,40 @@
...
+  method: 'put',
+  path: '/:hashId',
+  params: Joi.object().keys({
+    hashId: JoiHashId,
+  }).required(),
+  body: Joi.object().keys({
+    priorityLevel: Joi.number().integer().valid(1, 2).required()
+      .example('1'),
+    missedReports: Joi.number().integer().required().example('12'),
+    offlineForSeconds: Joi.number().integer().required().example('1800'),
```
everhardt, line 29: Typically we make the properties in the update body optional (so that the client doesn't have to provide the full object). Could you do that here as well?

Not wrong per se, but why is updating the deviceTypeHashId not allowed?

everhardt (changes requested):

src/routes/issue-trigger-rule/update.ts
Arinono: > Typically we make the properties in the update body optional (so that the client doesn't have to provide the full object).

I followed the REST spec, PUT needs all fields, PATCH doesn't. But PATCH is not allowed here

> Could you do that here as well?

Yes 👍 

> Not wrong per se, but why is updating the deviceTypeHashId not allowed?

I don't know yet how the creation of a rule at the `deviceType` level is gonna look in the monitoring environment as it was not really discussed. But I would expect a view where we list the `deviceTypes`, and when clicked we see its rules.
Not a list of rules where we can select the `deviceType`.
So updating the `deviceType` for a rule did not make sense to me (it would be a "transfer/copy" of rule really). But doing this opens a lot more questions:
Is there already a rule for this `deviceType` with the same `priority` ? Does the user want to override or abort ?
Is it a transfer or a copy/duplication of rule ?

src/routes/issue-trigger-rule/find.ts
```diff
@@ -0,0 +1,40 @@
...
+
+const controllerGeneratorOptions: ControllerGeneratorOptionsWithClient = {
+  method: 'get',
+  path: '/',
+  query: Joi.object().keys({
+    clientHashId: Joi.string().example('xd2rd4').required(),
+    deviceTypeHashId: Joi.string().example('xd2rd4'),
+  }),
+  right: { environment: 'READ' },
+  response: Joi.array().items(
```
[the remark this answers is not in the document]
Arinono: I forgot about pagination, I'll add it.

src/routes/issue-trigger-rule/update.ts
everhardt: What are your thoughts on this, @Arinono?

Arinono: @everhardt I like your suggestion. Actually this where I went with yesterday with my latest commit. But I need to run some extra tests to make sure I did it right

Arinono: 👆 https://github.com/withthegrid/platform-sdk/pull/126/commits/524886e40e035acdf8e02a7b974e5bea9bc9db7c#diff-917cd19227b7cd206abc3f72187da9de6deae23775aa2d8fd12735236279a230R7-R26

Arinono:
@everhardt I need to make sure the implementation is good on the platforms side, but it shouldn't prevent a re-review for the SDK

src/routes/issue-trigger-rule/update.ts
everhardt: And about the hashId vs. deviceType/priority discussion?

Arinono: oooh, maybe I miss understood something.
You suggest that a rule is "identified" by the deviceType/priority "combo" instead of hashId ?

If yes, I didn't consider/understood it because I never seen something like this before 🤔 

I'd have to look at some SQL docs to see/remember how to do it properly on the platform side

everhardt: > You suggest that a rule is "identified" by the deviceType/priority "combo" instead of hashId ?

Yes indeed, as we're not allowing more than 1 for the combination. A unique index on the two would be the solution in MySQL

Arinono: Ok, I'll try to make the changes by the end of the day (both sdk and platform) so you can have a look

Arinono: Last question, on this unique index (`UNIQUE KEY` ?), should I also include the `client_id` ?

> `UNIQUE KEY/INDEX rule (client_id, device_type_id, priority_level)` ?

everhardt: Yes :)!

everhardt (changes requested):

src/routes/issue-trigger-rule/find.ts
```diff
@@ -0,0 +1,50 @@
+import Joi from 'joi';
+import { EffectiveTableQuery, TableQuery, tableQuerySchemaGenerator } from '../../comms/table-controller';
+import { ControllerGeneratorOptionsWithClient } from '../../comms/controller';
+import { IssueTriggerRule, schema } from '../../models/issue-trigger-rule';
+
+type IssueTriggerRuleQueryFields = Omit<
+  Pick<IssueTriggerRule, 'priorityLevel' | 'deviceTypeHashId'>,
+  'priorityLevel'>;
```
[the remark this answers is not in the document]
Arinono: A (dumb) mistake

everhardt (approved):

everhardt:
Will hold off merging until the other PRs are also ready

everhardt (approved):
👍

PR #139, opened 2022-03-25
feat(pin): add pinGroupName to model

Arinono:
## Authors
@Arinono 

## Issues
withthegrid/platform-client#1002

Used by https://github.com/withthegrid/platform/pull/1111
Used by https://github.com/withthegrid/platform-client/pull/1038

## Implementation details

Adds a `pinGroup.name` to the model of a `pin`.

I chose to do 2 things:
- Not add more fields, because we don't need more atm
- Add the `name` in a separate object, next to `pinGroupHashId`:
  - Because I don't want to break anything using `pinGroupHashId`
  - Because I don't want to have a `pinGroupName` living with the rest of the object as it's not totally related

### DIFF
--- a/src/models/pin.ts
+++ b/src/models/pin.ts
@@ -4,9 +4,6 @@ import { schema as fieldsFromServerSchema, FieldsFromServer } from './fields/fie
 const schema = Joi.object().keys({
   hashId: Joi.string().required().example('e13d57'),
   pinGroupHashId: Joi.string().required().example('dao97'),
-  pinGroup: Joi.object().keys({
-    name: Joi.string().required().example('My Location'),
-  }).optional(),
   name: Joi.string().required().example('My port'),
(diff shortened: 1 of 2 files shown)


srieding (approved):

everhardt:
Two remark on this PR: 
* a return from the server should only contain optional properties if the data model requires them to be optional. This ensures that the client can rely on the data being provided and is not forced to deal with the weird situation where a pin does not have a related pin group name for example
* a model should only contain information from a single table. Otherwise it requires us to query multiple tables to deliver it, which defeats the purpose of it

Arinono:
@everhardt I moved the `pinGroup` part of the schema out of the model and into the response signature of the find endpoint.
I left it as `optional` even if it's technically impossible (i think), since deleting a pinGroup deletes the pins attached, but I prefer to be safe

PR #147, opened 2022-03-31
chore: update documentation link in readme

srieding:


docs/index.html
```diff
@@ -2249,7 +2249,7 @@
...
-<p>Programmatic access to the application requires a machine account. Almost everything a user can do in the user interface of the application, can also be done with a machine account. To obtain one, contact us at <a href="mailto:&#105;&#x6e;&#x66;&#x6f;&#64;&#x77;&#x69;&#x74;&#x68;&#x74;&#x68;&#101;&#103;&#114;&#x69;&#x64;&#x2e;&#99;&#x6f;&#x6d;">&#105;&#x6e;&#x66;&#x6f;&#64;&#x77;&#x69;&#x74;&#x68;&#x74;&#x68;&#101;&#103;&#114;&#x69;&#x64;&#x2e;&#99;&#x6f;&#x6d;</a>.
```
srieding, line 2252: It keeps changing this part from unicode hex to unicode decimal. Do you see why?

parkerjohn: Well that is pretty strange, I have no idea why it would do either conversion. This happens when you run the build?

srieding: yes it recognises the fact that it is an email adres and turns it into a mailto but during that process it esceapse everything and does so alternatingly between hex and decimal notations

parkerjohn (approved):

PR #151, opened 2022-04-05
chore: fix dependabot builds

srieding:


parkerjohn (approved):
We will have to figure out the hex conversion one day

PR #155, opened 2022-04-08
fix: only allow coordinates within the upper and lower bound of longi…

srieding:
…tude and latitude

refs: withthegrid/platform#1144


parkerjohn (approved):
Looks good, works perfectly!

PR #165, opened 2022-04-13
fix: command type not showing for dropped commands

srieding:
closes: withthegrid/platform-client#1057


src/models/analytics/analytics-tables.ts
```diff
@@ -429,6 +429,7 @@ const analyticsTables: Record<TableKeys, AnalyticsTable> = {
         level: 'Condition',
         deviceFields: 'Device form fields',
         deviceField: 'Device form field',
+        field: 'Form field',
```
srieding, line 432: I was creating an analytics query with pin and then a form field that gave a error in the console. Since it has not yet caused it's own sentry ticket I thought add it in here so it moves to production quickly.

parkerjohn:
Didn't we say we should add docs to .gitignore? Doesn't make sense to have it included in git and then generated again later on deploy anyway.

Issue #175, opened 2022-04-19
Git failed to push new release properly


parkerjohn:
The previous push to production failed to push some refs:

```Run npm run release

> @withthegrid/platform-sdk@17.42.7 release
> ts-node -P tsconfig.cjs.json release/

[[8](https://github.com/withthegrid/platform-sdk/runs/6074896939?check_suite_focus=true#step:5:8):52:40 AM] [semantic-release] › ℹ  Running semantic-release version 1[9](https://github.com/withthegrid/platform-sdk/runs/6074896939?check_suite_focus=true#step:5:9).0.2
[8:52:41 AM] [semantic-release] › ✔  Loaded plugin "verifyConditions" from "@semantic-release/changelog"
[8:52:41 AM] [semantic-release] › ✔  Loaded plugin "verifyConditions" from "@semantic-release/npm"
[8:52:41 AM] [semantic-release] › ✔  Loaded plugin "verifyConditions" from "@semantic-release/git"
[8:52:41 AM] [semantic-release] › ✔  Loaded plugin "verifyConditions" from "@semantic-release/github"
[8:52:41 AM] [semantic-release] › ✔  Loaded plugin "analyzeCommits" from "@semantic-release/commit-analyzer"
[8:52:41 AM] [semantic-release] › ✔  Loaded plugin "generateNotes" from "@semantic-release/release-notes-generator"
[8:52:41 AM] [semantic-release] › ✔  Loaded plugin "prepare" from "@semantic-release/changelog"
[8:52:41 AM] [semantic-release] › ✔  Loaded plugin "prepare" from "@semantic-release/npm"
[8:52:41 AM] [semantic-release] › ✔  Loaded plugin "prepare" from "@semantic-release/git"
[8:52:41 AM] [semantic-release] › ✔  Loaded plugin "publish" from "@semantic-release/npm"
[8:52:41 AM] [semantic-release] › ✔  Loaded plugin "publish" from "@semantic-release/github"
[8:52:41 AM] [semantic-release] › ✔  Loaded plugin "addChannel" from "@semantic-release/npm"
[8:52:41 AM] [semantic-release] › ✔  Loaded plugin "addChannel" from "@semantic-release/github"
[8:52:41 AM] [semantic-release] › ✔  Loaded plugin "success" from "@semantic-release/github"
[8:52:41 AM] [semantic-release] › ✔  Loaded plugin "fail" from "@semantic-release/github"
[...]

PR #176, opened 2022-04-19
fix: trigger a new release

parkerjohn:
fixes: #175

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,4 +5,4 @@ This package implements the API that allows communication with the withthegrid a
 See https://developer.withthegrid.com for documentation on this SDK.
 
 ## Questions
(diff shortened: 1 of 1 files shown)


jdbruijn:
FWIW you can also just do empty commits with the [`--allow-empty` flag](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---allow-empty) so you don't need random changes to trigger this 😉 If this occurs more often you can also look at the `workflow_dispatch` or `workflow_call` trigger for the GitHub Actions. I believe there is a trigger which you can use to just call on the main branch to run the workflow, and then of course update the docs in that scenario as well.

PR #200, opened 2022-05-10
fix: undefined response in error

everhardt:
part of withthegrid/platform-client#1156

### DIFF
--- a/src/comms/index.ts
+++ b/src/comms/index.ts
@@ -99,7 +99,7 @@ class Comms {
       if (Axios.isCancel(e)) {
         throw new CommsCanceledError();
       }
-      if (e.response !== undefined) {
+      if (typeof e.response === 'object' && e.response !== null) {
         if (e.response.status === 401) {
           throw new AuthenticationError(e.response.data.key, e.response.data.message);
         }


docs/index.html
```diff
@@ -2257,7 +2257,7 @@
...
-<p>Programmatic access to the application requires a machine account. Almost everything a user can do in the user interface of the application, can also be done with a machine account. To obtain one, contact us at <a href="mailto:&#105;&#110;&#x66;&#x6f;&#x40;&#119;&#x69;&#x74;&#x68;&#116;&#x68;&#x65;&#x67;&#x72;&#x69;&#100;&#46;&#99;&#111;&#109;">&#105;&#110;&#x66;&#x6f;&#x40;&#119;&#x69;&#x74;&#x68;&#116;&#x68;&#x65;&#x67;&#x72;&#x69;&#100;&#46;&#99;&#111;&#109;</a>.
+<p>Programmatic access to the application requires a machine account. Almost everything a user can do in the user interface of the application, can also be done with a machine account. To obtain one, contact us at <a href="mailto:&#x69;&#110;&#102;&#111;&#x40;&#x77;&#105;&#x74;&#x68;&#x74;&#x68;&#101;&#x67;&#x72;&#x69;&#100;&#46;&#x63;&#x6f;&#109;">&#x69;&#110;&#102;&#111;&#x40;&#x77;&#105;&#x74;&#x68;&#x74;&#x68;&#101;&#x67;&#x72;&#x69;&#100;&#46;&#x63;&#x6f;&#109;</a>.
```
everhardt, line 2260: @srieding Is there a way we can avoid these diffs?

parkerjohn: It seems like Github Pages expects to find the generated file in the main repo, so hard to avoid it.

srieding: Hopefully yes. But none came to mind immediately when we first saw this. I think it is building up momentum as a nuisance so I created: https://github.com/withthegrid/platform-sdk/issues/201

everhardt: I meant the back and forth indeed, thanks for filing the issue!

PR #213, opened 2022-05-19
chore(analytics-tables): refactoring, adding pluralization keys and helpers

Arinono:
## Authors
@arinono

## Issues
Refs withthegrid/platform-client#1169
Needed by <PR ID>

## Implementation details

Refactors analytics tables tests and add a couple to reach 100% coverage
Refactors parts of the analytics tables implementation, then adds pluralizations (keys and helpers)
Lastly adds a few tests for the l10n helpers.


src/models/translations.ts
```diff
@@ -1,18 +1,46 @@
...
+
+const isSimpleTranslation = (tr: Translation): tr is SimpleTranslation => typeof tr === 'string';
+
+const isPluralizedTranslation = (tr: Translation): tr is PluralizedTranslation => typeof tr === 'object'
+  && 'plural' in tr
+  && 'singular' in tr
+  && typeof tr.plural === 'string'
+  && typeof tr.singular === 'string';
+
+const localesKeys = ['en', 'nl'] as const;
```
srieding, line 30: you could reuse 
type Locale = keyof Translations;
from locale.ts

Arinono: Well, not really, `Locale` is based on `Translations`. So it would be a circular dependency.

That being said, I exported `localeKeys` and `LocaleKeys`, which are both unused outside, so no need for this, and `LocaleKeys` is the same as `Locale` so I will delete it.

parkerjohn (approved):
Works for me

PR #248, opened 2022-07-04
feat: removed inGrid comparison operator

yarkoyarok:
Part of withthegrid/platform#865

### DIFF
--- a/src/models/analytics-query.ts
+++ b/src/models/analytics-query.ts
@@ -94,7 +94,7 @@ type TimeGranularity =
   | 'dayOfTheWeek'
   | 'monthOfTheYear';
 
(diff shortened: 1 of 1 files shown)


srieding:
> As told [here](https://github.com/withthegrid/platform/issues/865#issuecomment-1173878782) it's breaking for some panels, but seems you are fixing that, so I hope it's already not breaking anything.

I think it is more than that. The comment in the ticket refers to the fact that we need to do migration because 1 existing panel will no longer be loadable if we go forward with this change. Here I was commenting on the fact that there are requests that are valid requests before this change and no longer are valid requests after this change. However if I look closer at the Joi schema I think that we will still accept them as we never restricted the values of comparison in the joi schema.

everhardt:
LGTM once https://github.com/withthegrid/platform/issues/865 is ready

PR #251, opened 2022-07-06
feat(commands): add deleteMulti endpoint types

Arinono:
## Authors
@Arinono 

## Issues
Refs withthegrid/platform-client#473
Used by withthegrid/platform#1364 and withthegrid/platform-client#1344

## Implementation details
I chose to create a new endpoint `DELETE /command/deleteMulti` for a few reasons.

First, if I updated the `DELETE /command` to support multiple `hashIds`, I think it would have brought a some complexity that is unnecessary (to the types, the documentation, ...)

Now, onto how it could have been done, well, I'm not sure, because no one knows for sure how it's done in a REST API. There is no standard for this operation, it can be:
- DELETE /command?hashid=one&hashid=two
- DELETE /command?hashids[]=one&hashids[]=two
- DELETE /command?hashids=one,two

And all of these are a breaking change.
(as a reminder, the current single delete endpoint is `DELETE /command/{hashId}`)

If I want to avoid a breaking change AND have a single endpoint, I would need an alternative schema (current OR new). And unless I missed smth, to support both, I would need an optional `hashId` in path or an optional `hashId` in the seachParams (no matter the shape). And doing this would mean that there is no required `hashId`, so it would need extra checks in the platform, and the validation happening in the SDK would be too loose.

### Screenshots

Here is how it look in the docs:
Screenshot 2022-07-08 at 14 38 27

### DIFF
--- a/src/routes/command/deleteMulti.ts
+++ b/src/routes/command/deleteMulti.ts
@@ -8,41 +8,48 @@ interface Request {
 }
 
 type ValidResponse = {
-	hashId: string
-	deleted: true
-}
+  hashId: string;
+  deleted: true;
+};
 
 type ErrorResponse = {
-	hashId: string
-	deleted: false
-	failureReason: string
-}
+  hashId: string;
+  deleted: false;
+  failureReason: string;
+};
 
 type Response = Array<ValidResponse | ErrorResponse>;
 
 const validResponseSchema = Joi.object().keys({
-	hashId: Joi.string().required().example('ga9741s'),
-	deleted: Joi.boolean().default(true).disallow(false).example(true),
+  hashId: Joi.string().required().example('ga9741s'),
+  deleted: Joi.boolean().default(true).disallow(false).example(true),
 });
 
 const errorResponseSchema = Joi.object().keys({
-	hashId: Joi.string().required().example('ga9741s'),
-	deleted: Joi.boolean().default(false).disallow(true).example(false),
-	failureReason: Joi.string().example('Device cannot be reached in time to cancel this command.'),
+  hashId: Joi.string().required().example('ga9741s'),
+  deleted: Joi.boolean().default(false).disallow(true).example(false),
+  failureReason: Joi.string().example(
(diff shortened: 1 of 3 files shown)


src/routes/command/deleteMulti.ts
```diff
@@ -0,0 +1,36 @@
...
+const controllerGeneratorOptions: ControllerGeneratorOptionsWithClientAndSupplier = {
+  method: 'delete',
+  path: '/',
+  body: Joi.object().keys({
+    hashIds: Joi.array().items(Joi.string()).required().example(['ga9741s']),
+  }).required(),
+  response: Joi.array().items(Joi.object().keys({
+    hashId: Joi.string().required().example('ga9741s'),
+    deleted: Joi.boolean().required().example(false),
+    reason: Joi.string().example('Device cannot be reached in time to cancel this command.'),
```
srieding, line 25: Will we ever send a reason with a deleted:true? So is {hashId: 'ga9741s', deleted: true, reason: 'you requested the deletion'} something the server will answer. You could consider naming the reason reasonFailure

srieding (approved):

src/routes/command/deleteMulti.ts
Arinono: Yes, I guess now that I'm sure this signature works, I could even go further and do a Union type, smth like:
```ts
type ValidResponse = {
  deleted: boolean = true
  hashId: string
}
type ErrorResponse = {
  deleted: boolean = false
  hashId: string
  failureReason: string
}
type Response = Array<ValidResponse | ErrorResponse>
```

(I don't remember if the boolean default is allowed in a type, but you get the idea I think)

And update the Joi schema accordingly

What do you think ?

srieding: I would like it better. But not a hard requirement

Arinono: 4cec2ac6b23c8cadf045c089a67f89cc30f852a5

everhardt:
This can be merged without my review!

PR #268, opened 2022-07-26
chore: dropped unused routes graph.getEdgeMeasurements and graph.getPinMeasurements

KZavorotna:
Is a part of https://github.com/withthegrid/platform/pull/1427

Closes https://github.com/withthegrid/platform/issues/1339


srieding (approved):

srieding:
I think it is fine to merge this. Are you sure the chore label will get you a version? If not then merging the one from platform will be difficult.

KZavorotna:
@srieding According to task, I removed unused routes.
Basically I can merge platform PR without this one and it still will work.

If this PR is merged first, there will be errors in platform one, because it will try to use routes, that don't exist in platform-sdk anymore.

Do I miss something here?
I think it would make sense to mark this one as the PR that closes the task.

srieding:
Yes, I think you should merge the platform first, keep this a chore and mark it as the one that closes the ticket

PR #269, opened 2022-07-28
feat: login a user over sso part 1 (logging in)

srieding:
refs: withthegrid/platform#1365
required by: https://github.com/withthegrid/platform/pull/1394


src/models/identity-provider-role.ts
```diff
@@ -0,0 +1,21 @@
...
+  rights: string[];
+}
+
+const schema = (): Joi.ObjectSchema => Joi.object().keys({
+  hashId: Joi.string().required().example('wasd2'),
+  identityProviderHashId: Joi.string().required().example('ds23d'),
+  clientHashId: Joi.string().example('as31db'),
+  supplierHashId: Joi.string().example('a2f1da'),
+  role: Joi.string().required(),
+  rights: Joi.array().items(Joi.string().example('READ')).required(),
```
Arinono, line 18: **Question**

Couldn't we have a more strongly typed `rights`/`roles` ?

You don't have to do anything here, but I would expect that we have a shared schema and (ts) type for these.

src/models/identity-provider.ts
```diff
@@ -0,0 +1,17 @@
...
+  name: string;
+  url: string;
+  ssoClientId: string;
+}
+
+const schema = (): Joi.ObjectSchema => Joi.object().keys({
+  hashId: Joi.string().required().example('wasd2'),
+  name: Joi.string().required(),
+  url: Joi.string().required().example('https://login.microsoftonline.com/abvsd23fasdfasdf/v2.0/.well-known/openid-configuration'),
+  ssoClientId: Joi.string().required().example('ed055c9d-c3b9-449a-818e-5ffd18d49f98'),
```
Arinono, line 14: **Request change**

I know it a development app, but I don't think it's a good idea 🌝 
```suggestion
  ssoClientId: Joi.string().required().example('e39cb59e-3a74-48cc-a71d-77b8ef037ec9'),
```

Arinono (approved):
LGTM 👌

src/models/identity-provider.ts
srieding: I would have sworn that I changed at leas some numbers and letters but it really is the same thing as on my db... shame on me...

src/models/identity-provider-role.ts
srieding: For roles -> no they will be defined by stedin and I definitely do not want to change our sdk if they decide to add a new role or change the spelling of one.
For rights yes but do you really want to? It is currently consistent with all other places where we define rights like https://github.com/withthegrid/platform-sdk/blob/22058adabb65be75b3db90734c0b3fb8f2fad6d1/src/models/user.ts#L14 And if we promise a certain number of acceptable values then introducing a new right would become a breaking change in the sdk.

Arinono: > And if we promise a certain number of acceptable values then introducing a new right would become a breaking change in the sdk

Very good point. I could argue that it doesn't have to be, but I don't have anything to add to verify this argument.
And I could also say that in some cases, having a change of rights **should be** a breaking change.

But none of the above is supported by proofs from me, and I like the first point: "It is currently consistent with all other places where we define rights"

Thanks for the answer 🙏

src/models/identity-provider.ts
Arinono: np, do you want me to remove my comment so that this is not visible anymore (you should rebase/rewrite the git history to remove the occurrence and force push) or it's not "important enough" (as it's a dev app) ?

srieding: not important enough. It is a clientId for an environment that has two users (just you and me) and is hosted by microsoft I'm sure they have rate limiting in place to make sure they can't be bugged too much by someone requesting something with this id that they are not authorized to obtain.

PR #276, opened 2022-08-02
feat: add theme to update and model

Arinono:
## Authors
@Arinono 

## Issues
refs withthegrid/platform#1324
Used by withthegrid/platform#1445
Used by withthegrid/platform-client#1392

## Implementation details

Very simple schema for the theme on the 2 kinds of environments.
I did account for `dark` theme even if it's mostly ignored/skipped everywhere. I just wanted to avoid breaking changes in the future if we want to add support for `dark` themes. And it seems pretty safe, since it can only be the same things as the `light` theme

### DIFF
--- a/src/commons/theme.ts
+++ b/src/commons/theme.ts
@@ -16,7 +16,6 @@ const colorsSchema = Joi.object().keys({
  * @internal
  */
 export const themeSchema = Joi.object().keys({
-  name: Joi.string().example('My new theme').required(),
   light: colorsSchema.required(),
   dark: colorsSchema.optional(),
 });
@@ -36,7 +35,6 @@ type Colors = {
  * @internal
  */
 export type Theme = {
-  name: string;
   light: Colors;
   dark?: Colors;
 };
--- a/src/commons/theme.ts
+++ b/src/commons/theme.ts
@@ -1,14 +1,15 @@
 import Joi from 'joi';
 
-const colorSchema = Joi.object().keys({
-  primary: Joi.string().required().example('#ff00ff'),
-  secondary: Joi.string().required().example('#ff00ff'),
-  accent: Joi.string().required().example('#ff00ff'),
-  error: Joi.string().required().example('#ff00ff'),
-  warning: Joi.string().required().example('#ff00ff'),
-  info: Joi.string().required().example('#ff00ff'),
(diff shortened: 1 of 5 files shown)


src/commons/theme.ts
```diff
@@ -0,0 +1,41 @@
+import Joi from 'joi';
+
+const colorSchema = Joi.object().keys({
+  primary: Joi.string().required().example('#ff00ff'),
```
srieding, line 4: We could consider adding in a .regex() checking against something like this ^#([A-Fa-f0-9]{6}|[A-Fa-f0-9]{3})$

srieding (approved):
LGTM

src/commons/theme.ts
Arinono: Haaaa, yes, I wanted to use `.hex()` but it was not exactly what I wanted. I forgot/was not aware of `.regexp()`, I'll do this change.
Thanks 🙏

src/commons/theme.ts
```diff
@@ -1,23 +1,24 @@
...
-const colorSchema = Joi.object().keys({
-  primary: Joi.string().required().example('#ff00ff'),
-  secondary: Joi.string().required().example('#ff00ff'),
-  accent: Joi.string().required().example('#ff00ff'),
-  error: Joi.string().required().example('#ff00ff'),
-  warning: Joi.string().required().example('#ff00ff'),
-  info: Joi.string().required().example('#ff00ff'),
-  success: Joi.string().required().example('#ff00ff'),
-  neutral: Joi.string().required().example('#ff00ff'),
+const colorSchema = Joi.string().required().example('#ff00ff').regex(/^#[0-9a-f]{6}$/i);
```
Arinono, line 3: I fixed the length at 6digit hex because on the client side, there is come computing on the hex values to generate the highlight colors for the dots in the map (when editing a line)

https://github.com/withthegrid/platform-client/blob/6d990dd56ab51211d4154a7bbf5835542bfdb96a/src/services/theme/index.ts#L63-L79

src/commons/theme.ts
Arinono: a3437c9d9be60d60aaa3f7273d657ebb18f3b469

PR #286, opened 2022-08-11
Fix platform client 1432 sort edges by proximity

GabrielPadureanu:
refs https://github.com/withthegrid/platform-client/issues/1432

is required by: 
https://github.com/withthegrid/platform-client/pull/1456
https://github.com/withthegrid/platform/pull/1467

- add a location query param and a sortBy option "proximity"
- display lines sorted by the distance to the location coordinates provided
image

### DIFF
--- a/src/routes/graph/find-edge.ts
+++ b/src/routes/graph/find-edge.ts
@@ -6,7 +6,7 @@ import { schema as edgeSchema, Edge } from '../../models/edge';
 import { TableQuery, EffectiveTableQuery, tableQuerySchemaGenerator } from '../../comms/table-controller';
 
 interface Query extends TableQuery {
-  location?: { latitude: number, longitude: number},
+  coordinates?: [number, number],
   includeDeleted?: boolean;
   boundingBox?: [number, number, number, number];
 }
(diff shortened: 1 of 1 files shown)


src/routes/graph/find-edge.ts
```diff
@@ -35,8 +37,12 @@ interface Response {
 const controllerGeneratorOptions: ControllerGeneratorOptionsWithClient = {
   method: 'get',
   path: '/edge',
-  query: tableQuerySchemaGenerator(Joi.string().valid('hashId', 'name').default('hashId'))
+  query: tableQuerySchemaGenerator(Joi.string().valid('hashId', 'name', 'proximity').default('hashId'))
     .keys({
+      location: Joi.object().keys({
```
srieding, line 42: I think this is a nice way to create the request. I think I might prefer it slightly over the current approach. However it is not consistent and I also do value consistency. Since it is unrealistic to change all locations where we already have coordinates for a slight preference I hope you would be willing to turn this implemenation into something closer to update-location.ts. The limits (Joi.number().min(-180).max(180).required(),) will be an improvement and the unnamed coordinates not so much but at least then we will be consistent with ourselves.

srieding (changes requested):

srieding (approved):

PR #287, opened 2022-08-12
feat(groups): added possibility to move all locations of group to another map later

KZavorotna:
Is required for https://github.com/withthegrid/platform-client/pull/1458

### DIFF
--- a/src/routes/graph/update-grid.ts
+++ b/src/routes/graph/update-grid.ts
@@ -13,6 +13,7 @@ interface Request {
     fields?: FieldsToServerUpdate;
     photo?: string | null;
     pinGroupHashIds?: string[]
+    mapLayer?: string;
   };
 }
 
@@ -35,6 +36,7 @@ const controllerGeneratorOptions: ControllerGeneratorOptionsWithClient = {
     fields: fieldsToServerUpdateSchema.example({ id: 'My grid' }),
     photo: Joi.string().allow(null).description('Should be a dataurl. Null clears the photo'),
     pinGroupHashIds: Joi.array().items(Joi.string()).description('Determines the set (and the order) of the pin groups in the grid'),
+    mapLayer: Joi.string().description('Determines a new map layer for all locations of group'),
   }).required(),
   response: Joi.object().keys({
     grid: gridSchema.required(),
--- a/src/routes/graph/update-grid.ts
+++ b/src/routes/graph/update-grid.ts
@@ -13,7 +13,6 @@ interface Request {
     fields?: FieldsToServerUpdate;
     photo?: string | null;
     pinGroupHashIds?: string[]
-    mapLayer?: string;
   };
 }
 
@@ -36,7 +35,6 @@ const controllerGeneratorOptions: ControllerGeneratorOptionsWithClient = {
     fields: fieldsToServerUpdateSchema.example({ id: 'My grid' }),
     photo: Joi.string().allow(null).description('Should be a dataurl. Null clears the photo'),
     pinGroupHashIds: Joi.array().items(Joi.string()).description('Determines the set (and the order) of the pin groups in the grid'),
-    mapLayer: Joi.string().description('Determines a new map layer for all locations of the group'),
   }).required(),
   response: Joi.object().keys({
     grid: gridSchema.required(),
--- a/src/routes/graph/update-grid.ts
+++ b/src/routes/graph/update-grid.ts
@@ -36,7 +36,7 @@ const controllerGeneratorOptions: ControllerGeneratorOptionsWithClient = {
     fields: fieldsToServerUpdateSchema.example({ id: 'My grid' }),
     photo: Joi.string().allow(null).description('Should be a dataurl. Null clears the photo'),
     pinGroupHashIds: Joi.array().items(Joi.string()).description('Determines the set (and the order) of the pin groups in the grid'),
-    mapLayer: Joi.string().description('Determines a new map layer for all locations of group'),
+    mapLayer: Joi.string().description('Determines a new map layer for all locations of the group'),
   }).required(),
   response: Joi.object().keys({
     grid: gridSchema.required(),
--- a/src/routes/graph/update-pin-groups.ts
+++ b/src/routes/graph/update-pin-groups.ts
@@ -17,11 +17,11 @@ const controllerGeneratorOptions: ControllerGeneratorOptionsWithClient = {
   method: 'post',
   path: '/update-pin-groups/:hashId',
   body: Joi.object().keys({
+    hashIds: Joi.array().items(Joi.string()).required(),
     mapLayer: Joi.object().keys({
       key: Joi.string().invalid('nodes'),
       includeEdges: Joi.boolean(),
     }),
-    gridHashIds: Joi.array().items(Joi.string()).description('PinGroups will be added at the end of the list in a grid'),
   }).required(),
   right: { environment: 'STATIC' },
   description: 'Updates pin groups',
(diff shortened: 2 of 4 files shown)


src/routes/graph/update-grid.ts
```diff
@@ -35,6 +36,7 @@ const controllerGeneratorOptions: ControllerGeneratorOptionsWithClient = {
     fields: fieldsToServerUpdateSchema.example({ id: 'My grid' }),
     photo: Joi.string().allow(null).description('Should be a dataurl. Null clears the photo'),
     pinGroupHashIds: Joi.array().items(Joi.string()).description('Determines the set (and the order) of the pin groups in the grid'),
+    mapLayer: Joi.string().description('Determines a new map layer for all locations of group'),
```
Arinono, line 39: I wouldn't make a suggestion if the other descriptions were not written like this, sorry 😓 

```suggestion
    mapLayer: Joi.string().description('Determines a new map layer for all locations of the group'),
```

everhardt:
Although `updateGrid` is the easiest place to put it, I think it's not really sound as we're not updating the grid, but only the underlying pinGroups. This becomes more ugly when we'd like to add a parameter `includeEdges: boolean`. Do you have a proposal how to deal with that?

KZavorotna:
@everhardt I see two options:
1) Update all pin groups one by one using existing `updatePinGroup` route
2) Create a route+controller that will accept a map layer, list of pin groups and `updateConnectedEdges` as boolean, that will update everything.

I also have two questions:
1) Should I move changing of map layer of pinGroup and edge to a same menu item and dialog as group for consistency? If yes, then this new controller can be used for doing this.
2) If we suggest to move connected edge with pin groups, should we suggest the same when moving edge? Like an option "move all connected pin groups".

everhardt:
On the options:
1. The upside is that there's almost no backend code that needs to be added (and/or repeated), but if you have 100's of locations in your group, it might take a while (esp. taking the rate limits into account). Also, there's repetition in backend processing (eg. `updateClientMbr` and `invalidateCache`)
2. That's probably fastest in execution, but requires custom code for every bulk action. I was also thinking of a variant of this option, that extends `updatePinGroup` with the possibility to provide multiple pinGroupHashIds (or a gridHashId). That would require a rewrite of the updatePinGroup route and would save on repetition, but it's way more difficult to keep performant (eg. updating the fields of many pinGroups would probably be pretty slow).

Backenders (@srieding @yarkoyarok @GabrielPadureanu) any suggestions?

On the questions:
1. I think it indeed makes sense to use that same dialog for single pinGroup changes as well!
2. That would be consistent indeed!

everhardt:
Okay, do you then agree that we should disallow properties of updatePinGroup that do not scale well to many pinGroups, like updating fields? And do you have a proposal for how to structure it in the API?

everhardt:
That makes sense to me. @srieding do you have any more feedback?

srieding:
I like the idea of a route update-multiple-pin-groups

KZavorotna:
@everhardt I have a question regarding bulk `update-multiple-pin-groups` route.
What should we allow to change except map layer?
For `update.pin-group` we have next:
`interface Request {
  params: {
    hashId: string;
  };
  body: {
    symbolKey?: string;
    geometry?: {
      type: 'Point';
      coordinates: [number, number];
    } | null;
    fields?: FieldsToServerUpdate;
    gridHashIds?: string[];
    gridName?: string;
    mapLayer?: string;
    deviceFields?: FieldsToServerUpdate;
    photo?: string | null;
  };
}`

Do I add only map layer for now and later we will add whatever else is needed?

everhardt:
I can imagine that `symbolKey` and `gridHashIds` would also be interesting, but propose to only do `mapLayer` right now to save time. I think `mapLayer` should have type `mapLayer?: { key: string, includeEdges: boolean }`

KZavorotna:
`grid.updatePinGroups` - sounds good.
Ok, then I'll start with map layer and I'll add symbolKey and gridHashIds if they don't take too much time.
Thanks everyone!

src/routes/graph/update-pin-groups.ts
```diff
@@ -0,0 +1,39 @@
+import Joi from 'joi';
+import { ControllerGeneratorOptionsWithClient } from '../../comms/controller';
+
+interface Request {
+  body: {
+    hashIds: string[];
+    symbolKey?: string;
+    gridHashIds?: string[];
```
Arinono, line 8: These fields are present here, but i don't think I saw the matching implementation in the platform. And given the messages (from Rob), I think this should be removed right ?

KZavorotna: You're right, updated it

src/routes/graph/update-pin-groups.ts
```diff
@@ -19,8 +17,6 @@ const controllerGeneratorOptions: ControllerGeneratorOptionsWithClient = {
   method: 'post',
   path: '/update-pin-groups/:hashId',
   body: Joi.object().keys({
-    hashIds: Joi.array().items(Joi.string()).required(),
```
Arinono: Hehe, the ordering was not the same here, you removed `hashIds` instead of `gridHashIds` (line 28, now 24)

Arinono (changes requested):

src/routes/graph/update-pin-groups.ts
KZavorotna: I should sleep more.
Updated

src/routes/graph/update-pin-groups.ts
```diff
@@ -0,0 +1,35 @@
...
+type Response = void;
+
+const controllerGeneratorOptions: ControllerGeneratorOptionsWithClient = {
+  method: 'post',
+  path: '/update-pin-groups/:hashId',
+  body: Joi.object().keys({
+    hashIds: Joi.array().items(Joi.string()).required(),
+    mapLayer: Joi.object().keys({
+      key: Joi.string().invalid('nodes'),
+      includeEdges: Joi.boolean(),
```
Arinono, line 23: Sorry I missed this, I think you need a `.default(false|true)` here ? (from what I've seen in other places in the codebase)

Arinono (approved):

src/routes/graph/update-pin-groups.ts
```diff
@@ -0,0 +1,35 @@
...
+      includeEdges?: boolean,
+    };
+  };
+}
+
+type Response = void;
+
+const controllerGeneratorOptions: ControllerGeneratorOptionsWithClient = {
+  method: 'post',
+  path: '/update-pin-groups/:hashId',
```
everhardt, line 18: * Should the `:hashId` be dropped in the path?
* We never use the `update`-verb (or `add` or `delete`) in the path, as it is already in the method.

Shall we therefore change the path to `/pin-groups`?

Arinono: > We never use the update-verb (or add or delete) in the path, as it is already in the method.

I'd agree with removing it, the HTTP method should be an indicator of the operation already. That being said, now that I'm saying this, I notice that it's a `POST` and not a `PUT` 😅. But it's a problem that we already have in the codebase. And making a breaking change in the platform for this is very likely the last thing that we want to do, so, not a problem for me here.

> Should the :hashId be dropped in the path?

Nice catch, it's unused in the platform, and doesn't make a lot of sense in this case.

KZavorotna: Updated

everhardt: 👍 

> Actually I think it should be a PATCH (partial update), but that's not the point

I agree. That's a mistake we made since the beginning, but consistently :).

PR #290, opened 2022-08-15
feat: unit property of quantities made optional

yarkoyarok:
Part of withthegrid/platform#809


src/models/quantity.ts
```diff
@@ -9,7 +9,8 @@ const schema = (apiVersion: number): Joi.ObjectSchema => Joi.object().keys({
   hashId: Joi.string().required().example('sajia1'),
   name: stringBeforeV7ElseStringOrTranslationSchema(apiVersion).required().example('Temperature'),
   color: Joi.string().required().example('#ff00ff'),
-  unit: Joi.string().required().example('K').description('Will be displayed with an SI-prefix (eg. k or M) if relevant'),
+  unit: Joi.string().allow(null).default(null).example('K')
```
srieding, line 12: This looks like a breaking change to me. Something that was guaranteed to have a string value now turns into something nullable. If the client's code can't handle that it might crash. I think we ought to send a '-' or a ' ' back in case there is no unit in the database and the apiVersion is lower than 8.

srieding (approved):

PR #316, opened 2022-09-09
feat(device types): added user facing identifier for devices

KZavorotna:
Is required for https://github.com/withthegrid/platform-client/pull/1545


src/routes/device/update-multiple-devices.ts
```diff
@@ -0,0 +1,29 @@
+import Joi from 'joi';
```
everhardt, line 1: * Shall we apply the same route naming as we use in https://github.com/withthegrid/platform-sdk/pull/287? So `updateDevices`?
* Taking https://github.com/withthegrid/platform-sdk/pull/287/files#r968069994 into account, shall we rename the route `path` to `/devices`?

everhardt (changes requested):
Looks good, 1 request.

src/routes/device/update-multiple-devices.ts
KZavorotna: Updated

everhardt (approved):
👍

Arinono (approved):
I didn't find anything out of the ordinary. Looking good! 👌

srieding:
No point in having a 4th pair of eyes look at this I would say

PR #348, opened 2022-10-14
fix: only part of conditierapport in issue hyperlinked

everhardt:
fixes withthegrid/platform#1643

### DIFF
--- a/src/commons/link-replacer.ts
+++ b/src/commons/link-replacer.ts
@@ -6,15 +6,15 @@ const command = ['command', 'opdracht'] as const;
 const matchers = [
   [
     'issues',
-    new RegExp(`(?<key>${issue.join('|')}) #(?<hashid>[\\w]{6,})`, 'gi'),
+    new RegExp(`(?<key>${issue.join('|')}) #(?<hashid>[a-z2-9]{6,})`, 'gi'),
   ],
   [
     'reports',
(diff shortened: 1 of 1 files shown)


src/commons/link-replacer.ts
```diff
@@ -1,20 +1,20 @@
...
-const report = ['condition report', 'report', 'rapport', 'meetrapport'] as const;
+const report = ['condition report', 'report', 'measurement report', 'rapport', 'meetrapport', 'conditierapport'] as const;
 const command = ['command', 'opdracht'] as const;
 
 // the regexp have been verified on https://makenowjust-labs.github.io/recheck
 const matchers = [
   [
     'issues',
-    new RegExp(`(?<key>${issue.join('|')}) #(?<hashid>[\\w\\d]{6,})`, 'gi'),
+    new RegExp(`(?<key>${issue.join('|')}) #(?<hashid>[\\w]{6,})`, 'gi'),
```
Arinono, line 9: Now that I think about this change, isn't `\w` too loose for our hashids ?
We don't have `_` in our hashids.

So it should be `[a-zA-Z\\d]` (or `[a-zA-Z0-9]` to make it simple, or even `[a-z0-9]` to be more strict) ?

everhardt: fair enough!

Arinono (approved):

PR #355, opened 2022-10-19
Feat platform 1619 add measurement threshold

GabrielPadureanu:
refs: https://github.com/withthegrid/platform/issues/1619
is required by: https://github.com/withthegrid/platform/pull/1655


srieding:
The closes comment should probably not be on the sdk because the sdk will only make it to production by merging the pr on platform. If you do it like this the issue in github gets closed when this is merged and if I am then very slow we only take this version of the sdk into production much later by merging the other pr.

src/routes/graph/get-edge.ts
```diff
@@ -18,8 +18,8 @@ interface Response {
   pins: Pin[];
   pinGroups: PinGroup[];
   nextReportBefore: Array<Date | null>;
-  thresholds: {
-    value: Threshold;
+  measurementThresholds: {
```
srieding, line 21: for something like this it is probably easiest to make no change at all

src/routes/quantity/find.ts
```diff
@@ -40,6 +40,7 @@ interface Response {
 const controllerGeneratorOptions: ControllerGeneratorOptionsWithClientAndSupplier = {
   method: 'get',
   path: '/quantities',
+  // rights to be updated to 'Threshold' which will be added soon
```
srieding, line 43: Make it a todo with the number of a ticket so if we ever forget we quickly can see when and where we forgot. Even better if in that ticket you also name this location because that makes the chance of forgetting much smaller

srieding (changes requested):

src/models/chart.ts
```diff
@@ -13,7 +13,7 @@ const schema = (apiVersion: number): Joi.ObjectSchema => Joi.object().keys({
     pinGroup: pinGroupSchema.required(),
     color: Joi.string().example('#ff0000').description('A hex color string').allow(null)
       .default(null),
-    pinThreshold: thresholdSchema.allow(null).required(),
+    pinThreshold: measurementThresholdSchema.allow(null).required(),
```
srieding, line 16: I think we could do without this change too?

srieding (changes requested):

src/models/connectivity-threshold.ts
```diff
@@ -0,0 +1,16 @@
+import Joi from 'joi';
```
srieding, line 1: Would have been nice to include a todo and a //deprecated into the IssueTriggerRule model. 
You could check out this review for inspiration: https://github.com/withthegrid/platform-sdk/pull/347

src/models/chart.ts
srieding: since this one is the type you can probably leave it

srieding (changes requested):

src/routes/measurement-threshold/find.ts
```diff
@@ -0,0 +1,111 @@
...
+  method: 'get',
+  path: '/',
+  query: tableQuerySchemaGenerator(Joi.string().valid('quantityName', 'pinGroupName', 'hashId').default('quantityName')).keys({
+    quantityHashId: Joi.string().allow(null).default(null),
+    pinGroupHashId: Joi.string().allow(null).default(null),
+    edgeHashId: Joi.string().allow(null).default(null),
+    pinHashId: Joi.string().allow(null).default(null),
+    type: Joi.string().valid('quantity', 'port'),
+  }),
+  // rights to be updated to 'Threshold' which will be added soon
```
srieding, line 69: please add the ticket number for the todo

src/routes/measurement-threshold/find.ts
```diff
@@ -66,8 +66,7 @@ const controllerGeneratorOptions: ControllerGeneratorOptionsWithClientAndSupplie
     pinHashId: Joi.string().allow(null).default(null),
     type: Joi.string().valid('quantity', 'port'),
   }),
-  // rights to be updated to 'Threshold' which will be added soon
-  right: { environment: 'ISSUES', supplier: 'READ' },
+  right: { environment: 'THRESHOLDS', supplier: 'READ' },
```
srieding, line 69: In the context of the general-information-markdown this makes no sense I think. What does it mean that for this route you need a read right from the supplier environment if we only know the ENVIRONMENT_ADMIN right for connectivity environments. If we need to access this from the connectivity environment this needs to be 
```supplier: 'ENVIRONMENT_ADMIN'```
However, at this point I think it is most likely that you don't need the supplier value here at all.

Comment (GabrielPadureanu) [reply src/routes/measurement-threshold/find.ts]: Yes, thanks for this

File: src/routes/measurement-threshold/find.ts (hunk 5 above)
Comment (GabrielPadureanu) [reply src/routes/measurement-threshold/find.ts]: Added the rights, no longer needed.

File: src/routes/quantity/find.ts (hunk 2 above)
Comment (GabrielPadureanu) [reply src/routes/quantity/find.ts]: Added the rights, no longer needed. Thanks

File: src/routes/measurement-threshold/find.ts [hunk 7]
```diff
@@ -0,0 +1,110 @@
...
+import { Quantity } from '../../models/quantity';
+import { stringBeforeV7ElseStringOrTranslationSchema } from '../../models/string-or-translations';
+import { schema as siNumberSchema, SiNumber } from '../../models/si-number';
+
+interface Query extends TableQuery {
+  quantityHashId?: string | null;
+  pinGroupHashId?: string | null;
+  edgeHashId?: string | null;
+  pinHashId?: string | null;
+  type?: 'quantity' | 'port';
```
everhardt, line 18: I meant putting them in the search string (so not listed explicitly here). That offers more flexibility (in line with the multiselect autocomplete of the front-end), for example to provide `quantity:aaa quantity:bbb` to get an OR for both.

src/routes/measurement-threshold/find.ts
```diff
@@ -0,0 +1,110 @@
...
+interface EffectiveRequest {
+  query: EffectiveQuery;
+}
+
+interface Threshold {
+  criticallyLow: SiNumber | null;
+  low: SiNumber | null;
+  high: SiNumber | null;
+  criticallyHigh: SiNumber | null;
+}
```
everhardt, line 42: Can't you just import src/models/threshold.ts? (And same for the schema below)

src/routes/measurement-threshold/find.ts
```diff
@@ -0,0 +1,110 @@
...
+    pinHashId: Joi.string().allow(null).default(null),
+    type: Joi.string().valid('quantity', 'port'),
+  }),
+  right: { environment: 'THRESHOLDS' },
+  response: (apiVersion: number): Joi.ObjectSchema => Joi.object().keys({
+    nextPageOffset: Joi.string().allow(null).example(null).required()
+      .description('This is the last page if nextPageOffset is null'),
+    rows: Joi.array().items(Joi.object().keys({
+      quantity: Joi.object().keys({
+        name: stringBeforeV7ElseStringOrTranslationSchema(apiVersion).required().example('Temperature'),
```
everhardt, line 75: as this route does not exist before v7, I think we can use [`stringOrTranslationSchema`](https://github.com/withthegrid/platform-sdk/blob/a61dfba3662c7a92e6e87257e1c6176b439baa2e/src/models/string-or-translations.ts#L5-L8) here?

src/routes/measurement-threshold/find.ts
```diff
@@ -0,0 +1,110 @@
...
+      }).allow(null),
+      threshold: Joi.object().keys({
+        criticallyLow: siNumberSchema.allow(null).required(),
+        low: siNumberSchema.allow(null).required(),
+        high: siNumberSchema.allow(null).required(),
+        criticallyHigh: siNumberSchema.allow(null).required(),
+      }).required(),
+    })).required(),
+  }),
+  description: 'Search through environment thresholds and measurement thresholds',
```
everhardt, line 100: I think `Search through measurement thresholds on environment and pin level` is more accurate?

everhardt (changes requested):
Looks good! Minor requests

src/routes/measurement-threshold/find.ts
GabrielPadureanu: Sounds good

src/routes/measurement-threshold/find.ts
GabrielPadureanu: Do you mean this [one](https://github.com/withthegrid/platform-sdk/pull/355/files#diff-2c5237690fe003707fb8b6741dceb16d49fc7defdf7dbc68f114cd0b6cb73200L16)? Maybe I can extend the MeasurementThreshold from the one above instead

src/routes/measurement-threshold/find.ts
GabrielPadureanu: Thanks, changing

src/routes/measurement-threshold/find.ts
everhardt: That's the one indeed. 

> Maybe I can extend the MeasurementThreshold from the one above instead

I would not derive a model from an interface in a controller, but other than that it would be possible. But I would be fine by using the whole `MeasurementThreshold` (so incl. `hashId` and `pinHashId`) here

GabrielPadureanu: Sorry, I meant to move it to the model and use it then in the controller. But by using the whole `MeasurementThreshold` would mean I also have to return those extra fields and the `threshold` object might come from a `quantity` and not have a `pinHashId`

everhardt: Ah, fair enough!

everhardt (approved):

PR #356, opened 2022-10-24
fix(quantity): add several name length limits

Arinono:
## Authors
@Arinono 

## Issue
Refs (and used by, kinda, not required, but better) withthegrid/platform-client#1407
Used by https://github.com/withthegrid/platform/pull/1907

While I was working on the ticket above: inputing very long names everywhere, I got an API error for a `quantity.unit` name being too long. DB schema has a `varchar(10)`, and the SDK doesn't check this.

So I'm opening a PR to add this to the SDK.

👀 ~As discussed with Rob, this won't be breaking change(s), since the API would have blocked the requests anyways.~
I introduced one 🤓 

### quantity.unit

Limiting to 10 chars to follow the DB schema
<details><summary>unit name longer than 10 chars</summary>

image

</details>

### Entities names 🚨 BREAKING 🚨 (stringOrTranslations)

The names of localisable entities were not limited in length.
So I updated the helper that was giving the schema for these to introduce a max length of `255` for version 9 (new version) onwards.

### measurement-filter

Add limit of `100` on name and `255` on description (add and update)

### webhook

Limited name to `255`

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -9,7 +9,7 @@ import * as commons from './commons';
 
 import { Response as MachineLoginResponse } from './routes/authentication/machine-login';
 
-const apiVersion = 8;
+const apiVersion = 9;
 
 class PlatformSdk {
   protected comms: Comms;
--- a/src/index.ts
+++ b/src/index.ts
@@ -9,7 +9,7 @@ import * as commons from './commons';
 
 import { Response as MachineLoginResponse } from './routes/authentication/machine-login';
 
-const apiVersion = 9;
+const apiVersion = 8;
 
 class PlatformSdk {
   protected comms: Comms;
--- a/src/models/string-or-translations.ts
+++ b/src/models/string-or-translations.ts
@@ -2,18 +2,28 @@ import Joi from 'joi';
 import { schema as translationsSchema, Translations } from './translations';
 import { Locale } from './locale';
 
-const schema = Joi.alternatives().try(
-  Joi.string().example('untranslated string').meta({ className: 'untranslatedString' }),
-  translationsSchema,
+const schema = (limit = 9999): Joi.AnySchema => Joi.alternatives().try(
+  Joi.string().example('untranslated string').meta({ className: 'untranslatedString' }).max(limit),
+  translationsSchema(limit),
 )
   .tag('stringOrTranslations')
   .meta({ className: 'stringOrTranslations' });
 
-const stringBeforeV7ElseStringOrTranslationSchema = (apiVersion: number): Joi.AnySchema => {
-  if (apiVersion <= 6) {
-    return Joi.string();
+const versionedStringOrStringOrTranslationSchema = (apiVersion: number): Joi.AnySchema => {
+  // [min, max, schema]
+  // range being [x;y)
+  const ranges: [number, number, Joi.AnySchema][] = [
+    [9, 9999, schema(255)],
+    [8, 9, schema()],
+    [0, 7, Joi.string()],
+  ];
+  for (let i = 0, len = ranges.length; i < len; i += 1) {
+    const [min, max, appliedSchema] = ranges[i];
+    if (apiVersion >= min && apiVersion < max) {
+      return appliedSchema;
+    }
   }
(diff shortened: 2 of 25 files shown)


srieding (commented):
Now that there is a migration for the database fields. Do we still stick to the same api version boundaries?

src/models/string-or-translations.ts
```diff
@@ -2,18 +2,39 @@ import Joi from 'joi';
...
+    translationsSchema(),
+  )
+    .tag('stringOrTranslations')
+    .meta({ className: 'stringOrTranslations' });
+};
 
-const stringBeforeV7ElseStringOrTranslationSchema = (apiVersion: number): Joi.AnySchema => {
-  if (apiVersion <= 6) {
-    return Joi.string();
+const versionedStringOrStringOrTranslationSchema = (apiVersion: number): Joi.AnySchema => {
```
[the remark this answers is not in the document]
Arinono: As discussed I went back to the simpler version of the function since there is no need to support localised names longer than 255 chars fb01c6e35b0bc43b29669518d2a0aa4370fe0317

src/models/string-or-translations.ts
```diff
@@ -2,18 +2,29 @@ import Joi from 'joi';
...
+    translationsSchema(),
+  )
+    .tag('stringOrTranslations')
+    .meta({ className: 'stringOrTranslations' });
+};
 
-const stringBeforeV7ElseStringOrTranslationSchema = (apiVersion: number): Joi.AnySchema => {
-  if (apiVersion <= 6) {
+const versionedStringOrStringOrTranslationSchema = (apiVersion: number): Joi.AnySchema => {
+  if (apiVersion < 9) {
```
srieding, line 24: This looks like a bad idea. Aren't we now breaking backward compatibility for anyone who was still on the stringBeforeV7 part?

src/routes/command-type/add.ts
```diff
@@ -26,7 +26,7 @@ const controllerGeneratorOptions: ControllerGeneratorOptionsWithSupplier = {
   method: 'post',
   path: '/',
   body: (apiVersion: number): Joi.ObjectSchema => Joi.object().keys({
-    name: stringOrTranslationsSchema.required().example('Measurement cycle'),
+    name: stringOrTranslationsSchema().required().example('Measurement cycle'),
```
srieding, line 29: If I read the code correctly this currently gives no limit for the name in the request. I thought limiting this in the request was the main reason we are even changing the sdk? Currenly you seem to mainly limit stuff in the models but then it does not fail as fast as it could?

srieding (changes requested):

src/routes/command-type/add.ts
Arinono: Indeed

src/models/string-or-translations.ts
Arinono: 1cd34ee

src/routes/command-type/add.ts
Arinono: 1cd34ee

src/index.ts
```diff
@@ -9,7 +9,7 @@ import * as commons from './commons';
 
 import { Response as MachineLoginResponse } from './routes/authentication/machine-login';
 
-const apiVersion = 8;
+const apiVersion = 9;
```
srieding, line 12: Is this still a correct change? If I understand the current state of the pull request correctly we are enforcing the stricter rules on add and update routes regardless of api version. That makes me wonder what breaking change the 9 version contains that the 8 version does not.

Arinono: Not being able to send localised names longer than 255 chars, it's still a breaking change I think

srieding: Yes it is breaking but it is breaking regardless of the apiVersion that you send it with. If someone sends us a long string with a apiVersion 4 then the server still validates against the length and gives an error.

srieding (approved):

src/index.ts
Arinono: 75937ee

PR #375, opened 2022-11-15
feat: prevent issues created

GabrielPadureanu:
refs: https://github.com/withthegrid/platform-client/issues/493

- add form field of `type`: `hashId` with possible `inputTypes`: `pinHashId` | `pinGroupHashId`
- allow base fields to be an object with a value (eg. can hold an hashId), and a value to be displayed
- add get device fields activity


src/models/fields/base-field-configuration.ts
```diff
@@ -178,6 +178,12 @@ const schema = (apiVersion: number): Joi.AlternativesSchema => {
       inputType: Joi.string().valid('file', 'files').required(),
       ...deprecatedValueOptions,
     }),
+    commonBaseFieldConfigurationSchema.keys({
+      type: Joi.string().valid('hashId').default('hashId').required(),
+      inputType: Joi.string().valid('pinGroupHashId', 'pinHashId').default('pinGroupHashId').required(),
+      ...deprecatedDefaultValue,
```
srieding, line 184: Why does a new field need deprecated values? Can't we just leave them out?

src/models/supplier-activities/get-device-fields.ts
```diff
@@ -0,0 +1,17 @@
...
+
+import { schemaConstructor as supplierActivityConstructor, SupplierActivity } from './base';
+
+interface GetDeviceFields extends SupplierActivity<'getDeviceFields'> {
+    triggerData: Record<string, never>;
+}
+
+const schema = supplierActivityConstructor(
+  'getDeviceFields',
+  Joi.object().keys({
```
srieding, line 11: how can one tell what object the deviceFields are for if this is the schema?

src/models/fields/base-field-configuration.ts
srieding: @everhardt maybe you help us here.

GabrielPadureanu: Doesn't work without it because the client sends a [defaultValue](https://github.com/withthegrid/platform-client/blob/main/src/views/generic/fields/field-configurations.vue#L227). Will check if it can be removed from the frontend without an impact.

I also see this comment:  [// can be dropped after deploy of related platform release](https://github.com/withthegrid/platform-sdk/blob/main/src/models/fields/base-field-configuration.ts#L46). I guess it was left behind and the related platform release happened?

GabrielPadureanu: > The from the client gets parsed by the platform-sdk models both on client and server. The client side parsing would remove it already I would expect? So if you drop the lines marked with `// can be dropped ...` and only update the server (not the client), I would still expect it to work.

That's true, it does if only the server has the sdk updated

> Also, the defaultValue you link to is for `type: 'string'` and `inputType: 'text'`. There `deprecatedDefaultValue` is not applied and [is still valid](https://github.com/withthegrid/platform-sdk/blob/a61dfba3662c7a92e6e87257e1c6176b439baa2e/src/models/fields/base-field-configuration.ts#L61).

Indeed, it is for `type: 'string'` and `inputType: 'text'`, but `add()` function is called from [here](https://github.com/withthegrid/platform-client/blob/main/src/views/monitoring-environment/settings/field-configurations.vue#L43) and sets `this.dialog.fieldConfiguration` acting as a default configuration which is overwritten, but if defaultValue: '' is not overwritten, it will remain set.

image

everhardt: Right, so the logic in the client would have to be updated to be able to drop those `deprecatedDefaultValue`'s. Let's leave that for another ticket then.

src/models/supplier-activities/get-device-fields.ts
GabrielPadureanu: Removed it as an activity

srieding (approved):

PR #376, opened 2022-11-17
feat(label): add get endpoint

Arinono:
## Authors
@Arinono 

## Issues
Refs withthegrid/platform-client#1737
Used by withthegrid/platform#
User by withthegrid/platform-client#


srieding (approved):
Nice the 'issueRoute'

PR #377, opened 2022-11-17
feat(mthresholds): return quantity unit

GabrielPadureanu:
refs: https://github.com/withthegrid/platform-client/issues/1370 and https://github.com/withthegrid/platform/issues/1619
is required by: https://github.com/withthegrid/platform/pull/1753
-  return quantity unit
<details><summary> response object example </summary>

image

</details>


srieding (approved):
You know how to get this deployed right? It looks safe to me so you may.

PR #378, opened 2022-11-21
feat(imports): add imports models and endpoints

Arinono:
## Authors
@Arinono @srieding 

## Issues
refs withthegrid/platform#1431

### Improvements on controller rights
I made the rights typed instead of a simple string as I added the right `IMPORTS`.
The reasons are:
- Type safety
- Intellisense/autocomplete

### Imports models
It should be pretty straightforward when it comes to `imports` types. I just made extensive use of Sum type for clarity and domain modelling.
The only thing I'm not sure about is the `Template` type, both regarding:
- place
- name

### Imports routes/endpoints

Most of the endpoints are very close to what we are used to.

The only spicy thing is the `B64XLSXFile` type and (joi) schema.
I made it following the patterns found, for example, for fields update for files: using base64 dataUrls.
The only difference is that it only accepts a mime-type for `.xlsx` files.


src/routes/imports/generate-template.ts
```diff
@@ -0,0 +1,41 @@
...
+      Joi.string().required().example('xd2rd4'),
+    )
+      .required()
+      .description('A list of location HashIDs'),
+    reportTypeHashIds: Joi.array().items(
+      Joi.string().required().example('xd2rd4'),
+    )
+      .required()
+      .description('A list of report type HashIDs'),
+  }).required(),
```
Arinono, line 28: todo: defined rules for min/max length for both arrays

Arinono:
@everhardt @srieding 
What do you think about these models and routes ?

I don't need a _final_ review, as maybe there are going to change a little bit with the client or platform implementation. But it's more about the general idea.

src/models/imports.ts
```diff
@@ -0,0 +1,126 @@
...
+}
+
+/**
+  * The processing of the import failed on one or more rows.
+  * There is a link available for the provided XLSX file where failed rows
+  * have an explanation of the error.
+  */
+type ErroredImport = BaseImport & {
+  state: 'errored'
+  link: string
```
srieding, line 41: I think it would be more consisten with eg file-from-server to call this url.

src/routes/imports/add.ts
```diff
@@ -0,0 +1,50 @@
+import Joi from 'joi';
+import { ControllerGeneratorOptionsWithClient } from '../../comms/controller';
+import {
+  ProcessingImport,
+  processingImportSchema,
+  InvalidImport,
+  invalidImportSchema,
+} from '../../models/imports';
+import { B64XLSXFile, b64XLSXFile } from './types';
```
everhardt, line 9: What's the added benefit of this schema and type? Why not use [FileToServer](https://github.com/withthegrid/platform-sdk/blob/1791b662a3629a7d626cbb90eb602f8572899fe5/src/models/file-to-server.ts)?

Applies to other occurrences as well

src/routes/imports/add.ts
```diff
@@ -0,0 +1,50 @@
...
+
+type Request = {
+  body: {
+    dataUrl: B64XLSXFile;
+    /**
+      * name override for the import.
+      * Defaults to filename
+      */
+    name?: string | null;
```
everhardt, line 18: 1. Why do you want to make it optional?
2. Why are you allowed to pass `null` if we already can leave it out?

src/models/imports.ts
```diff
@@ -0,0 +1,126 @@
...
+  | ErroredImport
+  | SuccessfulImport
+
+const templateSchema = Joi.object().keys({
+  pinGroupHashIds: Joi.array().items(Joi.string().required().example('xd2rd4')).required(),
+  reportTypeHashIds: Joi.array().items(Joi.string().required().example('xd2rd4')).required(),
+});
+
+const baseImportSchema = Joi.object().keys({
+  hashId: Joi.string().required().example('xd2rd4'),
```
everhardt, line 66: Convention currently is to make up a new hashId for a new model and use that consistently.

src/models/index.ts
```diff
@@ -21,6 +21,7 @@ import * as fieldToServerFull from './fields/field-to-server-full';
 import * as fileFromServer from './file-from-server';
 import * as fileToServer from './file-to-server';
 import * as grid from './grid';
+import * as imports from './imports';
```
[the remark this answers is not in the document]
Arinono: because `import` is a reserved word. So it was the easiest for now.
I don't know how to deal with this, I can:
- prefix (e.g. `$import`)
- rename, but to what ?

src/routes/imports/add.ts
Arinono: 2. True, I can remove this

1. The idea would be that the platform set the name following some rules, for example, defaults to the filename. But if the user wants a specific name, they can choose one.

src/models/imports.ts
Arinono: Does the convention defines a way to "pick" the hashId ?

Does it increment ? (if yes, how can I find the latest number)
or 
Should I just pick a random hashId and look if it's used somewhere ?

src/models/index.ts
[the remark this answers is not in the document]
everhardt: Right! Maybe we should call it an `importRequest`, similar to `exportRequest`?

src/routes/imports/add.ts
Arinono: None really, I'll replace with `FileFromServer`. 
The only advantage is that it enforces the mime-type. But I could find a way to extend `FileFromServer` to do so too if it's wanted

src/routes/imports/add.ts
everhardt: The filename is not included in the base64 I think? So the client would have to extract it (like it does [here](https://github.com/withthegrid/platform-client/blob/c10f323df5cbf4fe57500dbcbdd8c34e13b977d4/src/views/generic/file-upload.vue#L217-L218)) and feed it as name parameter. I think that name should be `string` only..

src/models/index.ts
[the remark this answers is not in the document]
Arinono: Good!

I also notices that `exportRequest` was part of the settings route. But I figured that import was "detached" enough that I wanted to make a specific route for them.
Are we okay with this direction ?

src/models/imports.ts
everhardt: The latter ;)

src/routes/imports/add.ts
everhardt: I don't think it is needed. We currently have mimetype optional in uploads by the way, see https://github.com/withthegrid/platform/blob/318093f40c9154ff6e4f30d49fe44ad589ec4b9f/src/services/files/files.ts#L42-L46.

src/models/index.ts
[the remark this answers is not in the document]
everhardt: Yes! Export should have gotten its own route, but it's still there because of historical reasons. That triggers me: downloads are personal in the sense that you can only see and access the downloads you created yourself. I'm not sure that is also desired for uploads: it might be interesting to see what others are doing as well. In that case, a username (and hashId) would also be interesting to know..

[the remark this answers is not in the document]
Arinono: > downloads are personal in the sense that you can only see and access the downloads you created yourself. I'm not sure that is also desired for uploads: it might be interesting to see what others are doing as well.

I'm not sure for these files. Data imports, imo, are shared in to everyone in the environment. So I'd say here, no restriction.

But adding the username/hashid is still valuable if we want to restrict later, and also to know would did the import in the first place.
So I think it's valuable to add this as `createdBy: { name, hashid }` ?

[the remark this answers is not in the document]
everhardt: I agree! The userHashId can go in the importRequest model but the userName has to be a property of the controller responses (as it comes from a different table in the database)

src/routes/import-request/add.ts
```diff
@@ -0,0 +1,34 @@
...
+
+const controllerGeneratorOptions: ControllerGeneratorOptionsWithClient = {
+  method: 'post',
+  path: '/',
+  body: fileToServer
+    .options({ presence: 'required' })
+    .required(),
+  right: { environment: 'IMPORT' },
+  response:
+    importRequestSchema,
```
everhardt, line 25: this should have a `.required()` behind it I think?

src/routes/import-request/add.ts
```diff
@@ -0,0 +1,34 @@
...
+
+type Response = (ImportRequest) & {
+  createdByUsername: User['name'];
+};
+
+const controllerGeneratorOptions: ControllerGeneratorOptionsWithClient = {
+  method: 'post',
+  path: '/',
+  body: fileToServer
+    .options({ presence: 'required' })
```
everhardt, line 21: What does this do? It's a new phenomenon..

src/routes/import-request/add.ts
```diff
@@ -0,0 +1,34 @@
...
+  importRequestSchema,
+} from '../../models/import-request';
+
+type Request = {
+  body: Required<FileToServer>;
+}
+
+type Response = (ImportRequest) & {
+  createdByUsername: User['name'];
+};
```
everhardt, line 15: This does not match the Joi schema. In other routes we don't extend models, but use different object properties, see eg. https://github.com/withthegrid/platform-sdk/blob/88852b1e88474d9511f91faa9ab1984f326579a6/src/routes/command/get.ts#L27-L29.

Also occurs at other routes

src/routes/import-request/delete.ts
```diff
@@ -0,0 +1,31 @@
...
+
+const controllerGeneratorOptions: ControllerGeneratorOptionsWithClient = {
+  method: 'delete',
+  path: '/:hashId',
+  params: Joi.object().keys({
+    hashId: Joi
+      .string()
+      .required()
+      .example('5x2znek')
+      .description('The HashID of the import to be deleted'),
```
everhardt, line 20: HashID -> hashId (multiple occurences)

src/routes/import-request/find.ts
```diff
@@ -0,0 +1,52 @@
...
+} | undefined;
+
+type EffectiveQuery = EffectiveTableQuery;
+
+interface EffectiveRequest {
+  query: EffectiveQuery;
+}
+
+interface ResponseRow {
+  import: ImportRequest & {
```
everhardt, line 20: See remark at .add route.

src/routes/import-request/get.ts
```diff
@@ -0,0 +1,36 @@
...
+  path: '/:hashId',
+  params: Joi.object().keys({
+    hashId: Joi
+      .string()
+      .required()
+      .example('5x2znek')
+      .description('The HashID for the import to be retrieved'),
+  }).required(),
+  right: { environment: 'IMPORT' },
+  response: importRequestSchema,
```
everhardt, line 27: Should have a `.required()` behind it I think

everhardt (changes requested):
Few things, sorry I did not spot them in the previous review!

src/routes/import-request/add.ts
srieding: should always be there. indeed.

src/routes/import-request/add.ts
srieding: It makes the name in the fileToServer required by changing the default for the keys within fileToServer. It saves us handling the possibility of an empty name in the controller.

src/routes/import-request/add.ts
srieding: check

src/routes/import-request/delete.ts
srieding: check

src/routes/import-request/find.ts
srieding: check

src/routes/import-request/get.ts
srieding: check

src/routes/import-request/add.ts
everhardt: I'm not sure that works properly: based on [the docs](https://joi.dev/api/?v=17.7.0#general-usage) I'd say it's equal to the `.required()` that's also already on the next line, so just requiring that the object is present.`.keys({ name: Joi.string().example('red-dot.png').description('The filename including extension').required() })` could work though..

everhardt (approved):

PR #410, opened 2022-12-21
Timescale

everhardt:


src/routes/chart/get.ts
```diff
@@ -17,8 +17,11 @@ interface Response {
   series: {
     pinHashId: string;
     quantityHashId: string;
-    linkHashId: string;
-    measurementsSummary: {time: Date, value: number}[];
+    measurementsSummary: {
+      time: Date;
+      value: number;
+      linkHashId: string;
```
everhardt, line 23: a) What's the reason for including it here instead of one level higher up?
b) What's the reason for including it at all?

everhardt: c) if we include it, it should have `a1b2c3` as example instead of `po177`.

everhardt:
I've reset the `timescale` branch to main. It's commits are now included in https://github.com/withthegrid/platform-sdk/pull/411

PR #411, opened 2022-12-21
feat: platform 1812 timescale remove measurement hashId

everhardt:


src/routes/chart/get.ts
```diff
@@ -19,7 +17,11 @@ interface Response {
   series: {
     pinHashId: string;
     quantityHashId: string;
-    measurements: Measurement[];
+    measurementsSummary: {
```
everhardt, line 20: a) What's the reason for including it here instead of one level higher up?
b) What's the reason for including it at all?
c) if we include it, it should have a1b2c3 as example instead of po177.

everhardt:
I propose to make use of the fact that a measurement is uniquely defined by pinHashId, quantityHashId and generatedAt. I propose to _not_ expose timeseriesHashId in the API.

PR #412, opened 2022-12-23
fix(measurement imports): user experience enhancements

srieding:
required by: https://github.com/withthegrid/platform-client/pull/1830
required by: https://github.com/withthegrid/platform/pull/1846

refs: withthegrid/platform-client#1822


everhardt (commented):
At itself fine, but isn't it easier to solve this in the front-end and let the SDK only handle pinGroups? That way it's also easier to adhere to the 200 limit.

PR #442, opened 2023-01-24
feat(feedback): obtain feedback from users

KZavorotna:
Is required for https://github.com/withthegrid/platform-client/pull/1921


src/routes/settings/feedback.ts
```diff
@@ -0,0 +1,29 @@
...
+}
+
+type Response = void;
+
+const controllerGeneratorOptions: ControllerGeneratorOptionsWithoutClientOrSupplier = {
+  method: 'post',
+  path: '/',
+  right: {},
+  body: Joi.object().keys({
+    theme: Joi.string().example('Email theme'),
```
Arinono, line 18: **question**

I don't see anything wrong with this, but why not make the theme mandatory too ?

src/routes/settings/feedback.ts
```diff
@@ -0,0 +1,29 @@
+import Joi from 'joi';
+import { ControllerGeneratorOptionsWithoutClientOrSupplier } from '../../comms/controller';
+
+interface Request {
+  body: {
+    theme: string;
```
Arinono, line 6: **requested change**

I think you need to change it to this to match the Joi schema below:
```suggestion
    theme?: string;
```

Arinono (changes requested):

Arinono:
One thing I thought about during my client review: 

> I just thought about this after the fact (this applied for both here and the SDK), but we might want to add length limit to both fields. I'm pretty sure our users won't try to spam our emails with dumps of text, but better be safe than sorry

https://github.com/withthegrid/platform-client/pull/1921#discussion_r1086665425

Arinono (approved):
LGTM! 👌

PR #451, opened 2023-02-02
Feat platform client 1865 clean up todo s

KZavorotna:
Is required for https://github.com/withthegrid/platform-client/pull/1949

PR #460, opened 2023-02-13
fix(add-pin): match TS request with Joi schema

Arinono:
## authors
@Arinono 

## issues
refs #451  [comment](https://github.com/withthegrid/platform-sdk/pull/451#discussion_r1103141845)


srieding:
Yes this is fine of course but it also at least needs a change here as far as I can tell: https://github.com/withthegrid/platform/blob/b4dc5bfb30dfb47913b97c6a2beec9822193d25e/src/controllers/graph/add-pin.ts#L63

Arinono:
I created the PR: https://github.com/withthegrid/platform/pull/1981

PR #471, opened 2023-02-22
feat(environment settings): added possibility to upload custom logo for monitoring environment

KZavorotna:
Is required for https://github.com/withthegrid/platform-client/pull/2033

### DIFF
--- a/src/routes/environment/get.ts
+++ b/src/routes/environment/get.ts
@@ -13,7 +13,8 @@ interface Request {
 interface Response {
   environment: Environment;
   environmentRights: string[];
-  userEnvironmentSettings: UserEnvironmentSettings,
+  userEnvironmentSettings: UserEnvironmentSettings;
+  environmentLogo: string | null;
 }
 
 const controllerGeneratorOptions: ControllerGeneratorOptionsWithoutClientOrSupplier = {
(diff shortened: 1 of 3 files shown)


src/routes/environment/get.ts
```diff
@@ -13,7 +13,8 @@ interface Request {
 interface Response {
   environment: Environment;
   environmentRights: string[];
-  userEnvironmentSettings: UserEnvironmentSettings,
+  userEnvironmentSettings: UserEnvironmentSettings;
+  environmentLogo: string | null;
```
Arinono, line 17: **requested change, I think**

I would have imagined that the logo was part of the `Environment` model as:
```ts
// ...
logo?: FileFromServer | null,
// ...
```

Arinono (changes requested):

Arinono:
**requested change**

Also the connectivity environment doesn't have the same changes. I'd imagine that the 2 kinds of environments want a customisable logo.

src/routes/environment/get.ts
Arinono: It was not about needing this info, but how the file is related to the environment, see https://github.com/withthegrid/platform/pull/2011#issuecomment-1453628465

PR #477, opened 2023-03-20
fix(chart.getPanels): make the query params optional

Arinono:
## authors
@Arinono 

## issues
refs withthegrid/platform-client#2151

used by https://github.com/withthegrid/platform-client/pull/2152
used by https://github.com/withthegrid/platform-sdk-private/pull/281
used by https://github.com/withthegrid/platform/pull/2081

## impl details

When the array is empty, the SDK strips the query param from the query, so the schema doesn't type check

### DIFF
--- a/src/routes/chart/get-panels.ts
+++ b/src/routes/chart/get-panels.ts
@@ -13,7 +13,7 @@ const kindsResources = ['pinGroups', 'pins'] as const;
 type KindsResources = typeof kindsResources[number];
 
 type Request = {
-  query: Record<KindsHashIds, Array<HashId>>;
+  query: Partial<Record<KindsHashIds, Array<HashId>>>;
 }
 
 type Panel = {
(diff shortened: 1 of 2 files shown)


src/routes/chart/update-panels.ts
```diff
@@ -22,10 +22,10 @@ type Panel = {
 };
 
 type Request = {
-  body: Record<KindsResources, Array<{
+  body: Partial<Record<KindsResources, Array<{
```
NickdeK, line 25: I'm not sure if this is what we want; how can the api update panels if the contents of the request body is optional?

src/routes/chart/get-panels.ts
```diff
@@ -13,7 +13,7 @@ const kindsResources = ['pinGroups', 'pins'] as const;
 type KindsResources = typeof kindsResources[number];
 
 type Request = {
-  query: Record<KindsHashIds, Array<HashId>>;
+  query: Partial<Record<KindsHashIds, Array<HashId>>>;
```
NickdeK, line 16: Talked about this on the chat with @Arinono where I voiced my concerns what we would expect if the query is optional. In this case it would mean it would return all panels, that could be fine?

srieding: I think it will mean that we give back nothing.

src/routes/chart/update-panels.ts
srieding: here it will update nothing, I think

